PR #159, opened 2023-03-21
Feat: open the content JSON on GitHub

katspaugh:
After editing a page, it will now open that page's content on GitHub, so that you can paste the new content.

I also had to run prettier on the components folder, so there are many unrelated changes, sorry.


DiogoSoaress (approved):
Nice and clean!

PR #162, opened 2023-03-23
fix: migrate Notion link

iamacook:
This updates the "Use our Hackathon Guide" link on the "Core" page to use our new Notion URL.

### DIFF
--- a/src/content/core.json
+++ b/src/content/core.json
@@ -207,7 +207,7 @@
     "buttons": [
       {
         "text": "Use our Hackathon Guide",
-        "href": "https://gnosis-safe.notion.site/Safe-d6c6ed61389041e28f5c7c925f653701",
+        "href": "https://safe-global.notion.site/Safe-d6c6ed61389041e28f5c7c925f653701",
         "variant": "button"
       },
       {


katspaugh (approved):

francovenica:
In which env I can see this? I tried https://safe-web-landing.staging.5afe.dev/core but is not there.

Also, in the comment you say that you have to change the "March for AA Hackaction" but you actually changed the "use our Hackaton guide" button, so which one is it?

iamacook:
> In which env I can see this? I tried https://safe-web-landing.staging.5afe.dev/core but is not there.

You will have to run the branch locally as this repo has no PR deployments.

> Also, in the comment you say that you have to change the "March for AA Hackaction" but you actually changed the "use our Hackaton guide" button, so which one is it?

Apologies, you are correct. It is the "Use our Hackathon Guide" link.

francovenica:
> You will have to run the branch locally as this repo has no PR deployments.

Is more trouble that is worth. For what I've seen is just a change in the link and nothing else. 
The link takes to an existing page and that's enough for me

PR #163, opened 2023-04-04
fix: Remove licenses page, update licenses link

usame-algan:
## What it solves

Part of https://github.com/safe-global/web-core/issues/1737

- Removes the licenses page, component, content
- Links to the new licenses page on app.safe.global


iamacook (approved):

francovenica:
So the ticket was kinda confusing, cuz the landing page in stg has different links.
It was explained to me that the old page is gone (in the case of staging is this one https://safe-web-landing.staging.5afe.dev/licenses) and the link will redirect to  https://app.safe.global/licenses

Is all good. Once this is deployed I'll check in the landing page in prod to make sure the links points to the correct licence page

Issue #171, opened 2023-04-20
Support the use of trailing slash


germartinez:
CLA scripts in all the Safe repositories point to `https://safe.global/cla/` (see https://github.com/safe-global/safe-homepage/blob/main/.github/workflows/cla.yml#L23 as an example).

https://safe.global/cla shows the right page, but https://safe.global/cla/ doesn't.
The same issue happens with all the other pages like `/core`, `/wallet`, etc.

Supporting the addition of a trailing slash would fix this issue.

katspaugh:
This can be probably fixed by adding a `pages/cla` folder with a symlink to `cla.tsx` inside.

katspaugh:
Fixed in #172.

tolbrino:
Still not working. `/cla` works but `/cla/` does not.

PR #172, opened 2023-04-20
Add Celo network

katspaugh:
Screenshot 2023-04-20 at 16 03 12


src/content/wallet.json
```diff
@@ -142,139 +142,76 @@
...
+          "src": "/images/chainsLogos/ArbitrumLogo.png",
+          "alt": "Arbitrum logo"
+        }
+      },
+      {
+        "name": "Celo",
+        "icon": {
+          "src": "/images/chainsLogos/CeloLogo.png",
+          "alt": "Celo logo"
+        }
```
usame-algan, line 213: Should we add the `"isNew": true` property here to display the "New" label?

usame-algan (approved):

src/pages/cla/index.tsx
usame-algan, line 1: Why do we need to put this page into a folder?

katspaugh: So that links like `/cla/` with a trailing slash work, see #171.

PR #181, opened 2023-04-24
feat: migrate careers to Ashby

### LINKED ISSUE #168: [Career page] Migrate Breezy to ATS
## What is the feature about

We are moving from Breezy to ATS and, as such, need update the Careers page (and header badge) to fetch listings from ATS.

The ATS setup will be ready by the end of April, to be migrated to by 2nd-4th May.

## The list of requirements

- Update [`useOpenPositions`](https://github.com/safe-global/safe-homepage/blob/main/src/hooks/useOpenPositions.ts) (and types) to fetch from ATS.
- Update any components that rely on the above accordingly.

## Links

- [Migration overview](https://www.notion.so/safe-global/ATS-migration-replace-API-fa9c469ffe1745c1894b3866e2b71bb8) (as well as contact person)

iamacook:
Resolves #168

Open positions are now loaded via Ashby.

The badge in the header and chip at the top of the "Careers" page render the count of open positions:

image

The open positions are listed correctly, linking the Ashby:

image

### DIFF
--- a/src/components/Careers/Positions/Positions.tsx
+++ b/src/components/Careers/Positions/Positions.tsx
@@ -10,15 +10,15 @@ import type { CardProps } from '@/components/Careers/Cards'
 
 import css from './styles.module.css'
 
-const parsePosition = ({ url, location, name }: Position): CardProps => {
+const parsePosition = ({ jobUrl, location, title }: Position): CardProps => {
   return {
-    title: name,
+    title,
     text: '',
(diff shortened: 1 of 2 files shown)


francovenica:
LGTM
I've cloned the repo, check the counter and the job positions. The request comes from ashby and it lists the same amount of open positions as the page (in this case 1)
Not much else to do since I cannot mess with the counter or add/remove open positions.
image

PR #186, opened 2023-05-02
fix: Don't cut off network chip New tag

### LINKED ISSUE #184: Allow `New` tag for networks in both rows
## Description

The New tag should be displayed for networks in both rows. Currently when enabling it for the top row it is being cut off because of the overflow hidden

image (4)

usame-algan:
## What it solves

Resolves #184 

Also refactors the network rows to remove hard-coded values and instead rely on css properties to restart animation from the correct spot. There should be no jumps anymore no matter how many logos are added or removed.

## How to test

1. Open the safe-homepage
2. Scroll down to Networks
3. Observe the `New` Tag is visible in both rows

## Screenshot

Screenshot 2023-05-02 at 12 44 31

### DIFF
--- a/src/components/Wallet/NetworkChip/index.tsx
+++ b/src/components/Wallet/NetworkChip/index.tsx
@@ -14,7 +14,7 @@ export type NetworkChipProps = {
 const NetworkChip = ({ name, icon, textColor, backgroundColor, isNew }: NetworkChipProps) => (
   <Box className={css.wrapper} sx={{ backgroundColor }}>
     <div className={css.icon}>
-      <img {...icon} />
+      <img {...icon} width={40} height={40} />
     </div>
     <Typography className={css.name} variant="body1" color={textColor}>
       {name}
--- a/src/components/common/Networks/index.tsx
+++ b/src/components/common/Networks/index.tsx
@@ -1,11 +1,12 @@
+import { shuffle } from 'lodash'
 import type { NetworkChipProps } from '@/components/Wallet/NetworkChip'
 import NetworkChip from '@/components/Wallet/NetworkChip'
-import { Box, Typography } from '@mui/material'
+import { Typography } from '@mui/material'
 import clsx from 'clsx'
 import css from './styles.module.css'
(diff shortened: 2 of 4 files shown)


src/components/Wallet/NetworkChip/index.tsx
```diff
@@ -14,7 +14,7 @@ export type NetworkChipProps = {
 const NetworkChip = ({ name, icon, textColor, backgroundColor, isNew }: NetworkChipProps) => (
   <Box className={css.wrapper} sx={{ backgroundColor }}>
     <div className={css.icon}>
-      <img {...icon} />
+      <img {...icon} width={40} height={40} />
```
iamacook, line 17: Is there a benefit to moving these from the CSS module?

src/components/common/Networks/index.tsx
```diff
@@ -43,15 +36,19 @@ const NetworksRow = ({
...
 }
 
 const Networks = ({ title, text, networks }: NetworksProps) => {
+  const [shuffledNetworks, setShuffledNetworks] = useState<NetworkChipProps[]>([])
   const chainsData = useContext(ChainsContext)
-  const shuffledNetworks = networks.slice(2, 8).reverse().concat(networks.slice(8).reverse(), networks.slice(0, 2))
+
+  useEffect(() => {
+    setShuffledNetworks(shuffle(networks))
+  }, [networks])
```
iamacook, line 51: As this is derived state, this initially confused me. Can you add a comment that this is to prevent hydration errors?

iamacook (approved):

src/components/Wallet/NetworkChip/index.tsx
usame-algan: Having the sizes inline prevents the Logos from blowing full size on page load if it takes a bit longer to parse the css. Its also generally recommended to have inline sizes for images.

src/components/common/Networks/index.tsx
usame-algan: I reverted back to how it was before

PR #187, opened 2023-05-02
Add Ecosystem page

### LINKED ISSUE #176: [Ecosystem page] Build a static page
Note: [database](https://github.com/safe-global/safe-homepage/issues/175) needs to be created first

## What is the feature about

Creating a new page to list all project built upon Safe.

## The list of requirements

1. Create static page that lists all project according to the design.

## ToDos

- [x] Add missing static page elements when ready from design
- [x] Fetch data and logos externally and remove `ecosystem-database.json` and `/images/ecosystem`

## Links

- [Epic](https://github.com/safe-global/safe-homepage/issues/161).
- [Figma design](https://www.figma.com/file/doceHL65jwlEbAGegLRnq2/Landing-Page?node-id=5050-10596&t=rMGDv6TmU6MfR1c6-0).

usame-algan:
## What it solves

Part of #176

## How it solves it

- Adds a new page "Ecosystem"
- Links to the page from the header navigation

## Screenshots

Screenshot 2023-05-02 at 17 28 55


src/components/Ecosystem/Projects/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                </InputAdornment>
+              ),
+            }}
+            fullWidth
+          />
+          <Typography mt={2}>
+            <Typography component="span" color="primary.light">
+              Example:
+            </Typography>{' '}
+            DeFi, DAO Tooling, Payments, NFT, Infrastructure
```
DiogoSoaress, line 122: What do you think of passing this string from the `content.json`?

DiogoSoaress (commented):
Looks neat!

One thing I noticed content related but that I think we should fix as it is breaking the mobile layout.
Screenshot 2023-05-04 at 09 16 31

Other remark is that the logo names should be capitalized as per one rule but we already talked about that and will be minded before merging into `main`

src/components/Ecosystem/Intro/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const EcosystemIntro = ({ title, text }: BaseBlock) => {
+  return (
+    <Container>
+      <Grid
+        container
+        className={css.container}
+        spacing={{ xs: 6, md: '30px', xl: '50px' }}
+        justifyContent="space-between"
+        alignItems="center"
```
iamacook, line 13: Nit: these can be moved to the `container`.

src/components/Ecosystem/Projects/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  primary_integration: string
+  packages: string
+  modules_guards: string
+  safe_apps_smart: string
+  networks: string
+  interface_can_you_create_a_safe: string
+  interface_can_you_import_an_existing_safe: string
+}
+
+const Card = ({ ...item }: EcosystemProject) => {
```
iamacook, line 43:
```suggestion
const Card = (item) => {
```

src/components/Ecosystem/Projects/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const categories = item.primary_category
+    .split(',')
+    .concat(item.secondary_categories.split(','))
```
iamacook, line 46: I'd suggest adding a `trim` in here to make sure all categories are sanitised, e.g.

```ts
.map((element) => element.trim())
```

src/components/Ecosystem/Projects/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+              return <Chip key={category + idx} className={css.categoryChip} label={category} />
+            })}
+        </div>
+      </div>
+    </Grid>
+  )
+}
+
+export const Projects = (): ReactElement => {
+  const projects = EcosystemDB as EcosystemProject[]
```
iamacook, line 99: What do you think about creating a guard for this? It could just check that every property in `EcosystemProject` exists and is a string.

src/components/Ecosystem/Projects/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+        <Grid item xs={12} md={9}>
+          <Grid container spacing="30px">
+            {projects &&
+              projects.map((ecosystemProject) => {
+                return <Card key={ecosystemProject.project} {...ecosystemProject} />
+              })}
+          </Grid>
+        </Grid>
+      </Grid>
+      <Divider sx={{ mt: '100px', mb: '100px' }} />
```
iamacook, line 141: Let's make a note to check this fixed spacing on mobile.

Nit:

```suggestion
      <Divider sx={{ my: '100px' }} />
```

iamacook (commented):
As always, amazing work! I only have a few small comments.

Did you optimise the images? We should resize those which are quite big as well, e.g. `avax.svg`. We could can a separate issue for this.

src/config/routes.ts
```diff
@@ -3,10 +3,13 @@ export const AppRoutes = {
...
   privacy: '/privacy',
-  licenses: '/licenses',
   index: '/',
   imprint: '/imprint',
+  ecosystem: '/ecosystem',
   core: '/core',
   cookie: '/cookie',
+  cla: {
+    index: '/cla',
+  },
```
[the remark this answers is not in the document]
usame-algan: Looks like we forgot to autogenerate the file last time the route was changed. I checked that we don't use `AppRoutes.cla` anywhere so there shouldn't be any issues.

src/components/Ecosystem/Projects/index.tsx
usame-algan: This is just a placeholder for now. Later on I assume these are part of the filtering e.g. when clicked they should filter the results by that category but something to discuss with Design.

src/components/Ecosystem/Projects/index.tsx
usame-algan: The `EcosystemProject` type is not finalized yet. I suggest we wait with this optimization until we have clarity regarding the final data structure.

usame-algan:
> Did you optimise the images? We should resize those which are quite big as well, e.g. avax.svg. We could can a separate issue for this.

Only the really bad ones (> 1mb). This is definitely something we should do before going to prod. I've added it as a discussion point in our meeting.

src/components/Ecosystem/Projects/index.tsx
DiogoSoaress: It looks to me like something that could be edited from GH. Your call

DiogoSoaress (approved):

src/components/Ecosystem/Projects/index.tsx
iamacook: > What do you think of passing this string from the `content.json`?

I've been looking at filtering and it makes sense to generate this from the database instead. I will include this there if it makes sense?

iamacook (approved):

src/components/Ecosystem/Projects/index.tsx
usame-algan: We should first clarify with Design if its intended to be interactive or not before implementing anything further. I left a comment in the Figma file.

iamacook: Ok, my proposal is we extract the categories (which needs to be done from the sidebar anyway) and select five random ones here. It needs not be dynamic but it aligns with the database if it is.

PR #188, opened 2023-05-03
fix: remove position count from search results

### LINKED ISSUE #180: Prevent open career count from appearing in search results
We are have an open position count in the header. This is also showing in search results. We need to prevent this from happening.

image

Discussion can be found [here](https://5afe.slack.com/archives/C04LENM73JQ/p1682327787113949).

iamacook:
This resolves #180 by removing the open position count from Google via the [`data-nosnippet`](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#data-nosnippet-attr) attribute.

image

### DIFF
--- a/src/components/common/Header/index.tsx
+++ b/src/components/common/Header/index.tsx
@@ -67,6 +67,12 @@ const Header = () => {
                   badgeContent={item.href === AppRoutes.careers ? positions.length : undefined}
                   color="primary"
                   className={css.badge}
+                  slotProps={{
+                    badge: {
+                      // @ts-expect-error - disable badge in search results
+                      'data-nosnippet': true,
+                    },
+                  }}
                 >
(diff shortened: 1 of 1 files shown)


src/components/common/Header/index.tsx
```diff
@@ -67,6 +67,12 @@ const Header = () => {
                   badgeContent={item.href === AppRoutes.careers ? positions.length : undefined}
                   color="primary"
                   className={css.badge}
+                  slotProps={{
+                    badge: {
+                      // @ts-expect-error - disable badge in search results
+                      'data-nosnippet': true,
```
usame-algan, line 73: Interesting find! TIL

usame-algan (approved):
Lets hope it works and Google picks it up 👍

PR #193, opened 2023-05-09
feat: reset filters button

iamacook:
This adds a button to reset the filters on the Ecosystem page:

image

image

### DIFF
--- a/src/components/Ecosystem/Projects/Projects.tsx
+++ b/src/components/Ecosystem/Projects/Projects.tsx
@@ -12,6 +12,7 @@ import {
   Toolbar,
   IconButton,
   Box,
+  Link,
 } from '@mui/material'
 import clsx from 'clsx'
 import { useMemo, useState } from 'react'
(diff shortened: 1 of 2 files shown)


src/components/Ecosystem/Projects/Projects.tsx
```diff
@@ -208,6 +218,11 @@ export const Projects = (): ReactElement => {
                 result{searchResults.length === 1 ? '' : 's'}
               </Typography>
             </Typography>
+            {!noFilters && (
+              <Link onClick={onResetFilters} className={css.reset} variant="caption">
+                RESET ALL
```
usame-algan: nit: You can also write "Reset all" and use css `text-transform: uppercase` to achieve the same result

usame-algan (approved):

src/components/Ecosystem/Projects/Projects.tsx
iamacook: `variant="caption"` does this already. I'll change it.

PR #199, opened 2023-05-10
feat: add pagination to search params

iamacook:
This adds the current page number of the Ecosystem page to the search params, also initialising based on the value.


usame-algan (commented):
In order for this to work we need to display the projects depending on the query param i.e. if I open the page with `?page=3` it should display projects 25-36.

usame-algan (approved):

PR #201, opened 2023-05-16
Fix: set the GA domain

### LINKED ISSUE #200: GA cookies are shared across all sub-domains (incl. app.safe.global)
It's unwanted from the privacy perspective that the cookies from safe.global as visible on app.safe.global.

## Solution

According to google:

> To limit the cookie only to the bare domain, you should set the domain parameter to be exactly the domain, without a leading dot. In your case, for the domain example.com, the cookie can be set as follows:

```js
document.cookie = "name=value; domain=example.com; path=/";
```

> This will make the cookie visible only on example.com and not on any subdomains. This is because according to the standard, if the domain parameter is specified without a leading dot, the cookie will only match the exact domain.

katspaugh:
Resolves #200.

These two options are probably redundant but I set both to err on the safe side.

### DIFF
--- a/src/hooks/useGa.ts
+++ b/src/hooks/useGa.ts
@@ -13,7 +13,7 @@ export const useGa = () => {
     if (IS_PRODUCTION && isAnalyticsEnabled) {
       ReactGA.initialize(GOOGLE_ANALYTICS_TRACKING_ID, {
         gaOptions: {
-          cookieFlags: `domain=${GOOGLE_ANALYTICS_TRACKING_ID};SameSite=Strict;Secure`,
+          cookieFlags: `SameSite=Strict;Secure`,
           cookieDomain: GOOGLE_ANALYTICS_DOMAIN,
         },
       })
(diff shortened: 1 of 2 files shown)


src/hooks/useGa.ts
```diff
@@ -11,7 +11,12 @@ export const useGa = () => {
   // Enable/disable tracking
   useEffect(() => {
     if (IS_PRODUCTION && isAnalyticsEnabled) {
-      ReactGA.initialize(GOOGLE_ANALYTICS_TRACKING_ID)
+      ReactGA.initialize(GOOGLE_ANALYTICS_TRACKING_ID, {
+        gaOptions: {
+          cookieFlags: `domain=${GOOGLE_ANALYTICS_TRACKING_ID};SameSite=Strict;Secure`,
```
usame-algan, line 16:
```suggestion
          cookieFlags: `SameSite=Strict;Secure`,
```
Having this option results in no cookies being set for me. `cookieDomain` sets the domain for the cookie already so I think we can remove it here.

usame-algan (approved):
👍

PR #204, opened 2023-05-23
fix: Ecosystem page issues

usame-algan:
## What it solves

Part of #176 

- Scales down the background particle on mobile
- Also searches within the project descriptions and categories, adjusts the search input placeholder accordingly
- `sanitize` now leaves in target attribute if present

## How to test

1. Open `/ecosystem`
2. Search for a project description or category and see it in the results
3. Click on links at the bottom of the page and observe them opening in a new tab


francovenica:
I was able to test it in my localhost just fine

The links opening in other tabs and the search also working with categories looks good

Not fully sold on the smooth scrollin, especially on mobile, where it could take more than half a second to write the next character if you don't actually know what you are looking for specifically.
My suggestion: give it more time, about 1.5 secs, and put a loading icon at the end of the bar so the user understand that the system is "doing something" and the scroll doesn't come as a surprise

I'll keep this ticket still in QA since other issues listed that must be discussed.

usame-algan:
> My suggestion: give it more time, about 1.5 secs, and put a loading icon at the end of the bar so the user understand that the system is "doing something" and the scroll doesn't come as a surprise

I've tested it with 1.5 sec but imo it feels too disconnected from the user action when the screen suddenly moves or interrupts scrolling. @TanyaEfremova what do you think about disabling smooth scroll on mobile and instead moving the three links above the search bar?

usame-algan:
After our sync on this I've now removed the smooth scroll and instead switched the search bar position with the links below so its closer to the result list ([Updated design](https://www.figma.com/file/doceHL65jwlEbAGegLRnq2/Landing-Page?type=design&node-id=5624-12833&t=ZoyC84iw1KPWYPbv-0))

src/components/Ecosystem/Projects/Projects.tsx
```diff
@@ -372,6 +354,12 @@ export const Projects = ({ items }: BaseBlock): ReactElement => {
                       </NextLink>
                     </Grid>
                   )}
+                  <Grid item xs={12}>
+                    <Typography variant="body2" maxWidth={470} mb={{ md: 8 }} mx="auto" textAlign="center">
+                      Listings are not endorsements and are only for informational purposes. Users should do their own
+                      research before interacting.
```
iamacook, line 360:
```suggestion
                      research before interacting with them.
```

iamacook (approved):

PR #206, opened 2023-06-13
fix: Support trailing slashes for all pages

usame-algan:
## What it solves

Supports trailing slashes for all pages e.g. opening `/cla` results in `/cla/`

## How to test

1. Open the homepage
2. Navigate to a few pages
3. Observe a trailing slash in the URL
4. Navigate to pages without a trailing slash via URL
5. Observe seeing the page

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -11,6 +11,7 @@ const nextConfig = {
       transform: '@mui/material/{{member}}',
     },
   },
+  trailingSlash: true,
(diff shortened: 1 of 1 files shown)


usame-algan:
Closing this PR because we don't want a trailing slash for every file since its not necessarily a directory and would result in the same content being visible on (technically) different URLs.

PR #207, opened 2023-06-19
fix: Search within project categories

usame-algan:
## What it solves

Searches within project categories that are persisted as an array instead of a comma-separated string which didn't work so well with fuzzy search

## How to test

1. Open the Ecosystem page
2. Search for Marketplace
3. Observe the same amount of results as if filtering by the same category


src/content/ecosystem.json
```diff
@@ -60,23 +60,23 @@
     ],
     "items": [
       {
-        "title": "Connect as a Safe App",
+        "title": "Connect as a Safe App.",
```
usame-algan, line 63: This was added as feedback [here](https://www.notion.so/safe-global/QA-b28438224f254a7da6d157803da0d9c9)

iamacook (changes requested):
As we are now adding the `categories_list`, we need not compute the `getProjectCategories` in other places: `_getFilteredProjects` and `ProjectCard`.

It may also be worth moving the addition of `categories_list` inside `useEcosystemData`. What do you think?

usame-algan:
> As we are now adding the `categories_list`, we need not compute the `getProjectCategories` in other places: `_getFilteredProjects` and `ProjectCard`.
> 
> It may also be worth moving the addition of `categories_list` inside `useEcosystemData`. What do you think?

Thanks, I've removed the other occurences of `getProjectCategories`. I would suggest we keep the `useEcosystemData` hook pure for now i.e. return whatever useSWR returns and do the adjustments in the component.

iamacook (approved):

PR #208, opened 2023-06-20
Add ecosystem meta image

usame-algan:
## What it solves

Adds a different og:image on the ecosystem page

## How to test

1. Open `/ecosystem`
2. Look for the ecosystem-image.jpg in the meta tags


src/content/ecosystem.json
```diff
@@ -1,6 +1,7 @@
 [
   {
     "pageTitle": "Safe – Ecosystem",
+    "image": "https://safe.global/images/ecosystem-image.jpg",
```
usame-algan, line 4: og:image doesn't work with relative urls so we have to use an absolute url here

PR #213, opened 2023-06-26
Chore: deploy all PR branches to dev

katspaugh:
We don't have per-branch deployments, so let's at least deploy all branches to dev. If multiple PRs are open at the same time, they will overwrite each other but that's OK.


usame-algan (approved):
Good idea! We haven't been using the dev deployment anyway.

PR #222, opened 2023-08-07
Back-merge `main` into `dev`

DiogoSoaress:
Sync `main` and `dev`


katspaugh (approved):

.github/workflows/deploy.yml
```diff
@@ -5,7 +5,6 @@ on:
 
   push:
     branches:
-      - dev
```
katspaugh, line 8: I would keep this if you're planning to use dev as your base branch for PRs, because then you can preview every branch on dev.

PR #226, opened 2023-08-10
feat: hook to ease scroll to an element

### LINKED ISSUE #227: [Whitepaper launching] Smooth scroll to the "Launching Protocol" section
When navigating to the Core page, smoothly slide in the "Launching Protocol" section to catch the user attention.

DiogoSoaress:
## What it solves

Resolves #227 

## How this PR fixes it

When opening the Core page, smoothly scrolls to the masthead section.
The target scroll position, ease and duration are reviewed by the design team.

### DIFF
--- a/src/components/Core/Masthead/index.tsx
+++ b/src/components/Core/Masthead/index.tsx
@@ -10,7 +10,7 @@ import Link from 'next/link'
 
 export const Masthead = ({ title, buttons, caption, image }: BaseBlock): ReactElement => {
   return (
-    <Container className={layoutCss.containerShort}>
+    <Container className={layoutCss.containerShort} id="masthead">
       <div className={css.container}>
         <Grid container spacing={{ xs: '90px', sm: '30px' }} justifyContent="space-between">
           <Grid item xs={12} md={7} lg={8}>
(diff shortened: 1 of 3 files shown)


katspaugh (commented):
Is this chatgpt-generated? Why are you not just using [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) with `behavior: 'smooth'`?

Also, on mobile it doesn't scroll all the way. The masthead is cut off at the bottom.

DiogoSoaress:
I tried first with the `behavior: 'smooth'` but I wanted a different ease function to give an acceleration effect.

>  on mobile it doesn't scroll all the way. The masthead is cut off at the bottom

It was requested to not scroll the masthead all the way to the top (that's why I'm subtracting a factor of `0.2 * window.innerHeight` to the masthead section top)

katspaugh (approved):
I'll approve but I would recommend looking into adding Framer Motion to this project as it's likely to have more custom animations in future.
And there's a plugin for Motion specifically for animated scrolling: https://github.com/breadadams/scroller-motion

DiogoSoaress:
Thanks for the references! I will definitely look into those as I agree the project will have more custom animations.

PR #230, opened 2023-08-16
feat: fetch Dune Safe numbers

### LINKED ISSUE #223: Dynamically fetch transaction volume, deployed safes and TVL
The numbers present in this Home section should be dynamically retrieved.

Screenshot 2023-08-08 at 09 40 12

DiogoSoaress:
## What it solves

Resolves #223 

## How this PR fixes it

- The Safe stats are pulled from Dune during build time.
Fallback numbers replace an unsuccessful fetch.

- On clicking the numbers in the Homepage, user is taken to the the respective Dune queries.

- Fetches: _totalTransactions_, _totalAssets_ and _totalSafesDeployed_.


katspaugh:
General note: don't think it's a good idea to make 3 API requests on each page load. I would suggest moving this to `getStaticProps` and updating the Dune numbers only during a build. With a steady release cadence they won't get outdated and it's obviously more convenient than editing them by hand.

src/pages/index.tsx
```diff
@@ -1,17 +1,26 @@
...
 
 const IndexPage: NextPage<InferGetStaticPropsType<typeof getStaticProps>> = (props) => {
   return <Home {...props} />
 }
 
 export async function getStaticProps() {
   const chainsData = await loadChainsData()
+  const totalTransactions = await fetchTotalTransactions()
+  const totalAssets = await fetchTotalAssets()
+  const totalSafesDeployed = await fetchTotalSafesDeployed()
```
katspaugh, line 14: You could also use Promise.all here.

katspaugh (approved):
I can't properly test on dev because the numbers are the same as the fallback, but the code looks good. Just one comment.

DiogoSoaress:
I've run the build locally to test it and it fetches 17M for `totalTransactions` (different from the fallback) so I think it is working well. We'll see now on staging after merging the branch into `main`

PR #232, opened 2023-08-22
fix: formatter now rounds numbers down and displays only 1 decimal

DiogoSoaress:
## What it solves

The displayed Safe stats should never be rounded up.

## How this PR fixes it

- Removes decimal part if number has more than 1 integer after the thousands order of magnitude formatting
- Truncates decimal part to one digit
- Use `en-US` locale for number formatting


iamacook:
You might be able to leverage the [`roundingMode` option](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#parameters) here.

DiogoSoaress:
I found that `roundingMode` is not supported yet in TypeScript https://github.com/microsoft/TypeScript/issues/43336#issuecomment-1092813558.

I've also checked the available libs in https://github.com/microsoft/TypeScript/tree/main/src/lib and the updates of Intl do not include NumberFormat so this option is unusable at the moment.

katspaugh (approved):

PR #243, opened 2023-10-06
feat: Governance page

DiogoSoaress:
## What it solves

Implementation of the Governance page according to [design](https://www.figma.com/file/doceHL65jwlEbAGegLRnq2/Landing-Page?type=design&node-id=6170-26660&mode=design&t=gSEvngP0jOJJuEsT-0).

### Obs
A list of delegates was provided by Andre, meaning we have no dynamic fetch in this iteration as the criteria to rank top delegates is still tbd.
Removed a bunch of unused assets. Mostly images that were replaced by parallax elements and were unused.

### Refactors (tiny)
- Add a margin-top to the Footer element instead of styling all pages last sections with margin-bottom
- Move ParallaxText from `components/Wallet` to the `components/common` folder
- Move TitleLogos from `components/Home` to the `components/common` folder


src/components/Governance/Delegates/delegates.json
```diff
@@ -0,0 +1,82 @@
...
+  {
+    "name": "0xAA",
+    "address": "0x3317AD9eDa6942b5a7BE5BA83346C0Ea82C3C26C",
+    "ens": "people-dao.eth",
+    "image": "0x3317AD9eDa6942b5a7BE5BA83346C0Ea82C3C26C_3x.png",
+    "reason": "I'm a core contributor at PeopleDAO. I value gnosis safe and advocate it actively.",
+    "contribution": "Wrote tutorials on gnosis safe for DAOs in English and Chinese.",
+    "twitter": "https://twitter.com/0xAA_Science"
+  }
+]
```
katspaugh, line 82: This is not a component, right? It's part of the content.

src/components/common/Header/index.tsx
```diff
@@ -31,8 +31,7 @@ const navItems: NavItemType[] = [
   },
   {
     label: 'Governance',
-    href: GOVERNANCE_LINK,
-    external: true,
+    href: AppRoutes.governance,
```
katspaugh, line 34: You might want to extract these links to a JSON as well at some point.

DiogoSoaress: The links we have in the `constants` file?

katspaugh: I mean the whole configuration of the header. What links and their names to display in the header.

DiogoSoaress: Good suggestion! I've moved that data to a content file as well.

katspaugh (approved):

PR #245, opened 2023-10-27
feat: rich text

DiogoSoaress:
## What it solves

Adds `richText` to the BaseBlock type. One step closer to how CMS store rich text.
Solves the need to have links in between text blocks.

## Screenshot

Screenshot 2023-10-27 at 15 34 00

### DIFF
--- a/src/components/Home/types.d.ts
+++ b/src/components/Home/types.d.ts
@@ -1,9 +1,12 @@
 import type { DetailedHTMLProps, ImgHTMLAttributes } from 'react'
 import type { ButtonProps } from '@mui/material'
 
+type RichTextVariant = 'text' | 'link'
+
 type BaseBlock = {
   title: string | JSX.Element
   text: string | JSX.Element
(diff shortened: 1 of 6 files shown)


katspaugh:
Good idea!
You could take it even further by allowing Markdown in all content types. Most md compilers allow allowlisting possible HTML tags, so it should be safe to insert as HTML.

DiogoSoaress:
Definitely! I will create an issue for that to implement it more generically.

I would however like to keep the scope of this PR minimal to allow embedding the link in the screenshot I shared (the new governance framework link). WDYT?

katspaugh (approved):

katspaugh:
I've approved it, but tbh this solution is pretty hacky.

DiogoSoaress:
Thanks 🙏 
I will refactor it when tackling https://github.com/safe-global/safe-homepage/issues/246

DiogoSoaress:
You're right. I'm closing this PR for https://github.com/safe-global/safe-homepage/pull/247

Content types markdown support will follow on https://github.com/safe-global/safe-homepage/issues/246

PR #248, opened 2023-10-30
fix: Update gov docs

DiogoSoaress:
- Updates Governance URLs


DiogoSoaress:
We still have to update links through code but CMS solution is underway!

katspaugh (approved):

PR #251, opened 2023-11-22
feat: Track "Social Login" campaign

DiogoSoaress:
## What it solves

Adds tracking to the "Social Login" campaign page according to the shared [doc](https://www.notion.so/safe-global/Tracking-events-Social-login-campaign-page-52c473bee00442e38f82159c1c20a8b9#e5705ffaff5242ae890ffac27a3e5735) 

## How this PR fixes it
- Creates an util function to fire a GA4 event
- Uses the tracking function when clicking the elements defined in the documentation

## Notes
I could only test the events were being received in our analytics dashboard by modifying `useGa.ts` locally. This hook initializes ReactGA4.

### DIFF
--- a/src/services/analytics/trackEvent.ts
+++ b/src/services/analytics/trackEvent.ts
@@ -1,0 +1,6 @@
+import ReactGA from 'react-ga4'
+import type { UaEventOptions } from 'react-ga4/types/ga4'
+
+export const trackEvent = (eventData: UaEventOptions | string) => {
+  ReactGA.event(eventData)
+}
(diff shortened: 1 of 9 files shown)


katspaugh (approved):
Looks good, much simpler than GTM.

src/services/analytics/trackEvent.ts
```diff
@@ -0,0 +1,6 @@
+import ReactGA from 'react-ga4'
+import type { UaEventOptions } from 'react-ga4/types/ga4'
+
+export const trackEvent = (eventData: UaEventOptions | string) => {
```
katspaugh, line 4: When can it be a string tho?

DiogoSoaress: Good catch. I've copied it over from the types definition file but we don't use it as a string ever. Removed it.

PR #254, opened 2023-12-04
feat: banner for SafeAuth

DiogoSoaress:
## What it solves

Modifies the Core page to show a banner allusive to the SafeAuth partnership

## Figma

https://www.figma.com/file/gxyAh1UOL6TXG1N5PWYLIB/Internal-Graphics?type=design&node-id=4317-257&mode=design&t=IKGFvK1GoeiU5V7S-4


src/components/Core/Masthead/index.tsx
```diff
@@ -7,12 +7,30 @@ import css from './styles.module.css'
...
+    sm: string
+    md: string
+    alt: string
+  }
+  backgroundImage: { sm: string; md: string }
+}): ReactElement => {
+  const isSmallScreen = useMediaQuery('(max-width:900px)')
+
+  const bgImage = isSmallScreen ? backgroundImage.sm : backgroundImage.md
+  const imageSrc = image ? (isSmallScreen ? image.sm : image.md) : undefined
```
DiogoSoaress, line 27: I am not sure this will be the closest implementation to what we'll have in the CMS. Open to hear other suggestions.

katspaugh: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#resolution_switching_different_sizes

katspaugh (approved):
I'll approve but those gradients are prety big. A quarter of a megabyte just for a background is excessive. You could easily make the same gradient with SVG or even CSS.

src/components/Core/Intro/index.tsx
```diff
@@ -3,11 +3,11 @@ import css from './styles.module.css'
...
 import { scrollToElement } from '@/lib/scrollSmooth'
 
-const CoreIntro = ({ title, text, link }: BaseBlock) => {
-  const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
+const CoreIntro = ({ title, text, link, scroll }: BaseBlock & { scroll?: { title: string; target: string } }) => {
+  const handleClick = (e: React.MouseEvent<HTMLAnchorElement>, target: string) => {
     e.preventDefault()
 
-    scrollToElement('#masthead', 200)
+    scrollToElement(target, 200)
```
katspaugh, line 10: Btw, the offset in this function can be replaced with CSS.
```css
      * {
        scroll-margin-top: 60px; // header height
      }
```

PR #256, opened 2023-12-07
feat: Recovery banner

DiogoSoaress:
## What it solves

Adds a banner to announce and link resources to the  RecoveryHub

## Figma

https://www.figma.com/file/Y3l2Y4UY6WsdOJrqKjN3J1/Web-Graphics?type=design&node-id=2927-109&mode=design&t=3bPDSZ6HRQhpfOPr-4


public/images/Wallet/buoy.png
katspaugh, line 1: Can you run it through tinypng again? Sometimes running a few times gives better results (with diminishing returns).

src/content/wallet.json
```diff
@@ -18,6 +18,25 @@
...
+    "buttons": [
+      {
+        "text": "Try now",
+        "href": "https://app.safe.global",
+        "variant": "button",
+        "color": "secondary"
+      },
+      {
+        "text": "Read more",
+        "href": "",
```
katspaugh, line 33: Dis tbd?

DiogoSoaress: Yes yes. Waiting for the blog URL

public/images/Wallet/buoy.png
DiogoSoaress: Oh that's a nice tip!
I've run it a couple times more but the spotlight in behind the image loses some quality (lines start to show) in exchange of 10KB. I'd think it's not worth in this case as it is more evident in the dark background. WDYT?

I will remember this for future png images tho. I feel we have many images that could be further optimized.

src/content/wallet.json
DiogoSoaress: Updated

katspaugh (approved):

PR #258, opened 2023-12-12
feat: Preserve search parameters in external links

DiogoSoaress:
## What it solves

We want to preserve the search parameters when navigating the homepage.

## How this PR fixes it

On page navigation the search parameters are merged and stored in the session storage key `SAFE__searchParams` and saved to a Context.
All links navigating to https://app.safe.global include the stored search parameters.


DiogoSoaress:
I would extend this behaviour to the other links in the homepage if you could navigate to a campaign page from https://safe.global , otherwise I don't see a justification to attach campaign related UTM codes when navigating to https://app.safe.global/

katspaugh:
* I would do it for all pages, not just campaigns
* Store the initial UTM params in the session storage so that they aren't gone on page refresh
* Why create a new component? You could just add it directly to ExternalLink.
* Next.js has a hook to get parsed params, which AFAIK is now preferred to useRouter, check out https://nextjs.org/docs/app/api-reference/functions/use-search-params

DiogoSoaress:
> I would do it for all pages, not just campaigns
Why?

> Store the initial UTM params in the session storage so that they aren't gone on page refresh
The search params are not lost on page refresh. This approach is however necessary if we want to hydrate the links on all pages with the UTM params, right?

katspaugh:
Because the user can:

1. come to a campaign page from a twitter link
2. navigate to another page on the site
2a. (potentially refresh the page)
3. only then go to app.safe.global

We lose the UTM params on steps 2 and 2a in your current implementation.

DiogoSoaress:
To clarify on the campaigns navigation:

* the user cannot reach the campaigns from safe.global and only the buttons in the campaign page link to the custom /welcome route in app.safe.global. The links on other pages link to https://app.safe.global, see https://safe.global/wallet for example.

- if you are in the current implementation with the UTM params shared on Twitter https://safe-web-landing.dev.5afe.dev/campaigns/social-login-gnosis?utm_source=t.co&utm_medium=cpc&utm_campaign=seedlessv1 and refresh the page, the search params are preserved.

Am I missing / overlooking something?

katspaugh:
Once again: the user can navigate to another internal page from the campaign page (e.g. via the links in the footer), at which point UTM params are lost.
I just don't see the point of limiting this solution to one particular type of pages while also making it less robust.

katspaugh:
I'll outline again my proposed solution which I shared before you started working on this PR:

* On page load, in a global context, save UTM params into the context and sessionStorage.
* Also attempt to restore any existing params from sessionStorage in case they are not in the URL.
* For all app.safe.global links, read UTMs from the global context and append them to the URL.

If you still disagree, please clarify the requiements with BI and marketing.

katspaugh (changes requested):
Importing the context, the append function and calling said function for every link in every component with external links isn’t very maintainable don’t you think? Please propose an alternative solution to discuss.

The context needs tests.

DiogoSoaress:
> Please propose an alternative solution to discuss.

I think having a HOC to call the context and append the search params is more maintainable, avoiding all the duplicated code.
You mentioned above to use the `ExternalLink` component but we don't have that component in this repo so I need to create a new component for this purpose. WDYT?

katspaugh:
I don't see the need for a HoC. Just create a custom link component, e.g. `SafeLink` that has all the logic and renders a Link.

src/components/common/SafeLink/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import Link from 'next/link'
+import { appendSearchParamsToURL } from '@/lib/appendSearchParamsToURL'
+
+const SafeLink = ({ href, children }: { href: string; children: ReactNode }) => {
+  const searchParams = useContext(SearchParamsContext)
+
+  const finalHref = appendSearchParamsToURL(href, searchParams)
+
+  return (
+    <Link href={finalHref} target="_blank" rel="noreferrer">
```
katspaugh, line 12: We should remove `rel="noreferrer"` for app.safe.global links.

src/components/common/SafeLink/index.tsx
```diff
@@ -0,0 +1,18 @@
+import SearchParamsContext from '@/contexts/SearchParamsContext'
+import { type ReactNode, useContext } from 'react'
+import Link from 'next/link'
+import { appendSearchParamsToURL } from '@/lib/appendSearchParamsToURL'
+
+const SafeLink = ({ href, children }: { href: string; children: ReactNode }) => {
+  const searchParams = useContext(SearchParamsContext)
+
+  const finalHref = appendSearchParamsToURL(href, searchParams)
```
katspaugh, line 9: `useMemo` perhaps?

katspaugh:
Looks good to me.

@usame-algan could you also take a look when you have a minute?

src/components/Campaign/Hero/styles.module.css
```diff
@@ -34,6 +34,10 @@
   margin: 0;
 }
 
+.textBlock a {
+  width: fit-content;
```
usame-algan, line 38: nit: Not sure if this is needed. The button looks the same to me with or without this css.

usame-algan (approved):
Looks good 👍 

I assume we use `SafeLink` for all links to `app.safe.global` but haven't checked each occurence.

src/components/Campaign/Hero/styles.module.css
DiogoSoaress: Screenshot 2023-12-14 at 16 26 43

vs 

Screenshot 2023-12-14 at 16 26 51

DiogoSoaress: The button keeps the width. It affects the link clickable area

DiogoSoaress:
> Launch Wallet button in the header it doesn't seem to be used

I've included it inside of the ContextProvider. Good catch @usame-algan

katspaugh (approved):
Thanks Diogo.
In future reviews we should keep in mind that we should always use SafeLink instead of Link for all links except when 100% internal.

DiogoSoaress:
Definitely!
I am very curious to see how carrying the UTM parameters will give another dimension to our website and wallet app traffic.

PR #261, opened 2024-01-02
fix: unique key prop warning in ecosystem project cards

DiogoSoaress:
## What it solves

- Moved key attribute to mapped result
- Adjusted the ProjectCard css to no overflow the category chips

Screenshot 2024-01-02 at 16 26 36
vs

Screenshot 2024-01-02 at 16 26 44


src/components/Ecosystem/Projects/ProjectCard.tsx
```diff
@@ -6,62 +6,60 @@ import GithubIcon from '@/public/images/github-icon.svg'
...
 
 export const ProjectCard = (item: EcosystemProjectWithCategories) => {
   const categories = item.categories_list
 
   return (
-    <div className={css.card}>
-      <Avatar className={css.image} src={ECOSYSTEM_DATA_URL + item.logo}>
-        &nbsp;
-      </Avatar>
+    <a href={item.project_website} target="_blank" rel="noreferrer">
```
usame-algan, line 15: In case there is no `item.project_website` the link is still there and "reloads" the page when the user clicks on it and there is now a warning that a cannot appear as a child of a. Any reason not to keep the conditional nested link like before?

src/components/Ecosystem/Projects/Projects.tsx
```diff
@@ -267,14 +267,14 @@ export const Projects = ({ items }: BaseBlock): ReactElement => {
               </Typography>{' '}
               {uniquePrimaryCategories.map((primaryCategory, idx, { length }) => {
                 return (
-                  <>
+                  <Fragment key={primaryCategory + idx}>
                     <SpecificCategoryFilter
                       key={primaryCategory + idx}
```
usame-algan, line 272: This key is not necessary anymore

src/components/Ecosystem/Projects/ProjectCard.tsx
DiogoSoaress: Oh thanks! I overlook the warning.
I'll attempt to revert to the previous method. The problem I faced was that the `<a>` element was positioned over the category Chips container impeding the scrolling.

usame-algan (approved):

PR #262, opened 2024-01-08
[DRAFT] feat: enable ISR

DiogoSoaress:


DiogoSoaress:
Other solutions than a Nextjs server are being discussed as a way to fetch live content.

PR #264, opened 2024-01-16
chore: increase package version

DiogoSoaress:
### Chores:
- Increase package version
- Rewording in the Careers page


katspaugh (approved):
Growing consistently is important indeed.

PR #265, opened 2024-01-18
feat: Blog

DiogoSoaress:
# What it solves
Implements:

- [Blog landing page](https://www.figma.com/file/CYPABu9F8t34m6hkMxQnDB/Blog?type=design&node-id=9501%3A1628&mode=design&t=vyth5U31FJmXxMVn-1)
- [Blog post](https://www.figma.com/file/CYPABu9F8t34m6hkMxQnDB/Blog?type=design&node-id=9710%3A1175&mode=design&t=vyth5U31FJmXxMVn-1)

# How this PR fixes it
Implements new route `/blog` to the blog landing page.
The blog's landing page content and articles are hosted in [Contentful headless CMS](https://www.contentful.com/).
I explain the current CMS `content` and `field types` in 
[Blog content types](https://www.notion.so/safe-global/Blog-content-types-e7561d1dd07e473086014fe209a2be30?pvs=4)

## Remarks
- Contentful does not provide a solution to solve circular references. In our case it happens that `relatedPosts` have posts that reference the currently open one which causes a serialization error (Circular references cannot be expressed in JSON)
- The search on the Ecosystem page is similar to the implemented in for the Blog landing page so I abstracted it to a common component. I suggest further refactors on the Projects page are tackled in a different PR.
- The Banner section is omitted until further clarification on what to do with the signed up emails.
- [ ] Port remaining [`safe.mirror.xyz`](https://safe.mirror.xyz/) articles. This will be coordinated together with the Content Writer and SEO expert and should not be a block to this PR.


package.json
```diff
@@ -29,6 +30,7 @@
     "contentful": "^10.6.1",
     "fuse.js": "^6.6.2",
     "isomorphic-dompurify": "^1.0.0",
+    "json-stringify-safe": "^5.0.1",
```
DiogoSoaress, line 33: This library replaces circular references with a placeholder string, allowing safe conversion to JSON.
I use it to solve circular references in the blog articles, where a post can reference another post that references back itself.

src/pages/blog/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  if (!blogHome || !postsEntries) {
+    return {
+      notFound: true,
+    }
+  }
+
+  // replaces circular dependencies allowing safe conversion to JSON
+  const featuredPost = JSON.parse(jsonStringifySafe(blogHome.fields.featured))
+  const mostPopular = JSON.parse(jsonStringifySafe(blogHome.fields.mostPopular))
+  const allPosts = JSON.parse(jsonStringifySafe(postsEntries.items))
```
katspaugh, line 31: I really don't get how you can have circular references in what already comes as JSON.

DiogoSoaress: I explained it in the PR description to add some more context to the reviewer. I've also included in the PR description the CMS [content types](https://www.notion.so/safe-global/Blog-content-types-e7561d1dd07e473086014fe209a2be30?pvs=4)

The error I get without using this safe serialization is
Screenshot 2024-02-05 at 10 00 53

katspaugh: Yes, that's when Next.js serializes the static props for SSG because it needs to [bundle the data as JSON](https://nextjs.org/docs/pages/building-your-application/data-fetching/get-static-props#statically-generates-both-html-and-json) with the web page. You could avoid passing that data as static props in the first place.

To be honest, I don't see the point of using a CMS for static site generation. You ideally want to avoid making a release every time  someone posts in the blog or edits a typo. So I would move all that data fetching into a client-side component and ony statically render the HTML layout.

DiogoSoaress: That makes a lot of sense. I will apply the suggested changes

DiogoSoaress: Moving the fetching to the client side has a drawback... In order to fetch content on the client-side we will have to initialize the client (https://www.contentful.com/developers/docs/javascript/sdks/), exposing:
	- `SPACE_ID` 
	- `ACCESS_TOKEN` - key to the read-only API https://www.contentful.com/developers/docs/references/content-delivery-api/ that allows consuming published content. IMO it does not expose any sensitive information. However, by including it in the public bundle someone can pick it (despite being used for a read-only API) and abuse the number of requests leading to hit the usage limits.


I still tried to initialize the Contentful client in the `getStaticProps` but the object is then not serializable (contains functions).

Given that, do you think it is preferred to use client-side fetching over SSG?

katspaugh: That's how it's normally done, yes. Some people prefer to proxy the API to hide the key and limit it to a specific domain but I don't think it's worth it.

katspaugh (approved):

PR #267, opened 2024-02-07
chore: do not index staging and dev deployments

DiogoSoaress:
Instruct search engine crawlers not to index deployments other that PRODUCTION.
Having `5afe.dev` pages indexed in search engines damages our SEO performance.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -39,7 +39,6 @@ jobs:
       - uses: ./.github/workflows/build
         with:
           secrets: ${{ toJSON(secrets) }}
-          is-production: false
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -39,7 +39,6 @@ jobs:
       - uses: ./.github/workflows/build
         with:
           secrets: ${{ toJSON(secrets) }}
-          is-production: false
```
DiogoSoaress, line 42: `is-production` input seems to be treated as a string rather than a boolean

katspaugh (approved):

PR #268, opened 2024-02-07
feat: dynamically fetch DAO voting stats

DiogoSoaress:
# What it solves
Replaces image assets by API calls to return
- number proposals
- number of unique delegates
- number of delegators
  
  Screenshot 2024-02-08 at 00 56 54

Fixes the Snapshot title parsing
Screenshot 2024-02-08 at 11 02 33

# How this PR fixes it
Fetches the voting numbers in `getStaticProps` similarly to what we do for [SafeStats](https://github.com/safe-global/safe-homepage/blob/main/src/hooks/useSafeStats.ts)  and shares them through a `VotingDelegationContext`


compojoom (approved):
nice 👍

PR #278, opened 2024-02-23
fix: change homepage meta title

DiogoSoaress:
Change homepage title as agreed for SEO reasons

### DIFF
--- a/src/content/home.json
+++ b/src/content/home.json
@@ -1,6 +1,6 @@
 [
   {
-    "pageTitle": "Safe — Leading the Smart Account standard on Ethereum",
+    "pageTitle": "Safe: Ethereum Smart Accounts",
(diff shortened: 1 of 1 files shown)


src/content/home.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "pageTitle": "Safe — Leading the Smart Account standard on Ethereum",
+    "pageTitle": "Safe: Ethereum Smart Accounts",
```
iamacook, line 3: All other pages start with "Safe —". Do we also want to change them?

iamacook (approved):

src/content/home.json
DiogoSoaress: Good point. I will raise this before merging the PR

PR #280, opened 2024-03-04
fix: display Youtube playlists in RichText

DiogoSoaress:
Enhance the `youtubeRegex` to accept playlists.

Can be tested in https://fix_embeded_entry--homepage.review.5afe.dev/blog/safe-conference-for-web3-berlin-2024

### DIFF
--- a/src/lib/urlPatterns.ts
+++ b/src/lib/urlPatterns.ts
@@ -1,5 +1,5 @@
 export function isYouTubeUrl(url: string) {
-  const youtubeRegex = /^(https?:\/\/)?(www\.)?(youtube\.com\/watch\?v=|youtu\.be\/)[\w-]{11}$/
+  const youtubeRegex = /^(https?:\/\/)?(www\.)?(youtube\.com\/watch\?v=|youtu\.be\/)[\w-]{11}(&list=\w+)?$/
   return youtubeRegex.test(url)
 }
 
--- a/src/lib/urlPatterns.ts
+++ b/src/lib/urlPatterns.ts
@@ -1,6 +1,7 @@
 export function isYouTubeUrl(url: string) {
   const youtubeRegex =
     /^(https?:\/\/)?(www\.)?(youtube\.com\/(watch\?v=|playlist\?)|youtu\.be\/)([\w-]{11})?(&?list=[\w]{34})?/
+
   return youtubeRegex.test(url)
 }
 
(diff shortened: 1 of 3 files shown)


src/lib/urlPatterns.ts
```diff
@@ -13,6 +14,24 @@ export function extractYouTubeVideoId(url: string) {
   return searchParams.get('v')
 }
 
+export function extractYouTubePlaylistParams(url: string) {
+  const urlObj = new URL(url)
+  const searchParams = new URLSearchParams(urlObj.search)
+  return {
+    list: searchParams.get('list'),
+    index: searchParams.get('index'),
+  }
```
iamacook, line 23:
```suggestion
  return Object.fromEntries(urlObj.searchParams.entries())
```

If changed, I'd rename the function to something like `extractYouTubeSearchParams`.

src/lib/urlPatterns.ts
```diff
@@ -13,6 +14,24 @@ export function extractYouTubeVideoId(url: string) {
...
+}
+
+export const getYoutubeVideoSrc = (url: string) => {
+  const videoId = extractYouTubeVideoId(url)
+  const { list, index } = extractYouTubePlaylistParams(url)
+
+  return `https://www.youtube-nocookie.com/embed/${videoId ? videoId : 'videoseries'}${list ? `?list=${list}` : ''}${
+    !videoId ? `&index=${index ?? 1}` : ''
+  }`
```
iamacook, line 32: These ternaries are a bit confusing to follow. It might be better to use the `URL` interface. What do you think?

```suggestion
  const urlObj = new URL(`https://www.youtube-nocookie.com/embed/${videoId ? videoId : 'videoseries'}`)
  if (!videoId) {
    urlObj.searchParams.set('index', index ?? 1)
  }
  if (list) {
    urlObj.searmParams.set('list', list)
  }
  return urlObj.toString()
```

src/lib/urlPatterns.ts
```diff
@@ -1,5 +1,6 @@
 export function isYouTubeUrl(url: string) {
-  const youtubeRegex = /^(https?:\/\/)?(www\.)?(youtube\.com\/watch\?v=|youtu\.be\/)[\w-]{11}$/
+  const youtubeRegex =
```
iamacook, line 2: Can we simplify this to match line 9?

iamacook (approved):
My suggestions are nits so I'll let you decide whether to merge it as it otherwise LGTM.

PR #296, opened 2024-03-27
fix: default to a total assets static value

DiogoSoaress:
Because Dune's total assets stored query is returning a wrong value we should use the fallback value instead of fetching.


usame-algan (approved):
👍

PR #300, opened 2024-04-09
feat: Token page

DiogoSoaress:
## What is solves

Token landing page 

## Figma 
https://www.figma.com/file/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?type=design&node-id=10347-1968&mode=design&t=DXjbuWmTS5i9RVsX-0


src/components/common/Page/index.tsx
```diff
@@ -0,0 +1,28 @@
+import getComponentByName from '@/lib/getComponentByName'
+import type { Entry } from 'contentful'
+import { type TypeLandingPageSkeleton } from '@/contentful/types'
+import { isEntryType, isEntryTypeBaseBlock } from '@/lib/typeGuards'
+import MetaTags from '@/components/common/MetaTagsContentful'
+
+export type LandingPageEntry = Entry<TypeLandingPageSkeleton, undefined, string>
+
+const Page = ({ landingPage }: { landingPage: LandingPageEntry }) => {
```
DiogoSoaress, line 9: Page is fully composed from the CMS `content` blocks so the landing page layout can be adjusted solely from there.

src/pages/[slug].tsx
Comment (DiogoSoaress) [inline src/pages/[slug].tsx:1]: Allows landing pages to be created from the CMS. You create a `landingPage` entry and define the slug of the new page.

src/styles/theme.ts
DiogoSoaress, line 1: Had to modify the theme Typography while https://github.com/safe-global/safe-homepage/pull/301 is not merged.

src/components/Pressroom/index.tsx
```diff
@@ -36,7 +37,7 @@ export type PressRoomProps = {
 const PressRoom = ({ pressRoom, allPosts, totalAssets }: PressRoomProps) => {
   const { metaTags, featured, numbers, investors, timeline, news, podcasts, videos } = pressRoom.fields
 
-  const numbersList = numbers.filter(isEntryTypeSimpleBaseBlock)
+  const numbersList = numbers.filter(isEntryTypeBaseBlock) || []
```
iamacook, line 40:
```suggestion
  const numbersList = numbers.filter(isEntryTypeBaseBlock)
```

src/components/Token/ButtonsWrapper/index.tsx
```diff
@@ -0,0 +1,34 @@
+import LinkButton from '@/components/common/LinkButton'
+import { Button } from '@mui/material'
+import SafeLink from '@/components/common/SafeLink'
+import { type ButtonEntry } from '@/config/types'
+import css from './styles.module.css'
+
+type ButtonsWrapperProps = { buttons: ButtonEntry[] }
+
+const ButtonsWrapper = ({ buttons }: ButtonsWrapperProps) => {
+  if (!buttons || buttons.length === 0) return null
```
iamacook, line 10: The type defines `buttons` as defined.

```suggestion
  if (buttons.length === 0) return null
```

src/components/Token/CenteredTitleCards/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <Container className={layoutCss.containerMedium}>
+      <div className={css.textBlockContainer}>
+        <RichText {...title} />
+
+        {text && <RichText {...text} />}
+      </div>
+
+      <Grid container columnSpacing="30px" rowGap="16px" mt={{ xs: '40px', md: '80px' }}>
+        {itemsList?.map((item, index) => (
```
iamacook, line 54:
```suggestion
        {itemsList.map((item, index) => (
```

src/components/Token/Floaters/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  direction: AnimationDirection
+  duration: number
+  speed: number
+}
+
+const Floater = ({ radius, angle, size, speed, direction }: FloaterProps) => {
+  const prefersReducedMotion = usePrefersReducedMotion()
+  const [currentAngle, setCurrentAngle] = useState(angle)
+  const requestRef = useRef<number>()
+  const speedMultiplierRef = useRef<number>(prefersReducedMotion ? 0.5 : 2)
```
iamacook, line 20: Why do we need to store this value in a ref?

src/components/Token/Floaters/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+        const adjustedSpeed = direction === 'reverse' ? -speed : speed
+        return (prevAngle + (adjustedSpeed * speedMultiplierRef.current * deltaTime) / 1000) % 360
+      })
+      requestRef.current = requestAnimationFrame(animate)
+    },
+    [speed, direction, speedMultiplierRef],
+  )
+
+  useEffect(() => {
+    speedMultiplierRef.current = prefersReducedMotion ? 0.5 : 2
```
iamacook, line 43: This is the same value as defined in the initial state. We could assign this to a reuseable variable.

src/components/Token/Floaters/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const animate = useCallback(
+    (time: number) => {
+      if (!lastFrameTime.current) {
+        lastFrameTime.current = time
+      }
+      const deltaTime = time - lastFrameTime.current
+      lastFrameTime.current = time
+      setCurrentAngle((prevAngle) => {
+        const adjustedSpeed = direction === 'reverse' ? -speed : speed
+        return (prevAngle + (adjustedSpeed * speedMultiplierRef.current * deltaTime) / 1000) % 360
```
iamacook, line 35: Nit: can you break this down a bit with comments? It's hard to follow.

src/components/Token/Hero/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { isAsset, isEntryTypeButton } from '@/lib/typeGuards'
+
+const Hero = (props: BaseBlockEntry) => {
+  const { caption, title, buttons, image } = props.fields
+
+  const buttonsList = buttons?.filter(isEntryTypeButton) || []
+
+  return (
+    <div className={css.heroSpacer}>
+      {/* Token Backgroud Gradients */}
```
iamacook, line 20:
```suggestion
      {/* Token Background Gradients */}
```

src/components/Token/Hero/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+          {/* Display All Floaters */}
+          {orbitRing.floaters.map((floater, floaterIndex) => (
+            <Floater
+              key={floaterIndex}
+              radius={orbitRing.diameter / 2}
+              angle={floater.angle}
+              size={floater.size}
+              direction={orbitRing.direction as AnimationDirection}
+              duration={orbitRing.duration}
+              speed={360 / orbitRing.duration + floater.speed / 70}
```
iamacook, line 72: Nit: the `radius` value is used twice here and `speed` is hard to follow. I'd suggest moving them to declaritive variables outside of the render, e.g.

```ts
{OrbitRingsList.map((orbitRing, orbitIndex) => {
  const radius = orbitRing.diameter / 2
  return (
    ...
```

src/components/Token/Hero/token-hero-rings.json
```diff
@@ -0,0 +1,110 @@
+[
+    {
+      "id": "Ring1",
+      "duration": 20,
+      "direction": "normal",
+      "diameter":300,
```
iamacook, line 6: You should run this file through a formatter. There are some formatting inconsistencies, e.g.

```suggestion
      "diameter": 300,
```

src/components/Token/Floaters/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+        duration: Math.random() * 0.5,
+      }}
+      style={{
+        width: `${size}px`,
+        height: `${size}px`,
+        borderRadius: '50%',
+        position: 'absolute',
+        left: `calc(50% + ${x}px - ${size / 2}px)`,
+        top: `calc(250px + ${y}px - ${size / 2}px)`,
+        background: '#12FF80',
```
iamacook, line 72: Can you use a palette var. here?

src/components/Token/ParallaxToken/ParallaxUseCasesElement.tsx
```diff
@@ -0,0 +1,44 @@
...
+const trailingShadow = (
+  <>
+    <div className={`${css.shadow} ${css.first}`} />
+    <div className={`${css.shadow} ${css.second}`} />
+  </>
+)
+
+const ParallaxGovernanceProcessElement = ({ items }: { items: Entry<TypeBaseBlockSkeleton, undefined, string>[] }) => {
+  const generateParallaxWrapper = (index: number, className: string) => {
+    const depth = [0, 1, 2] as const
```
iamacook, line 19: This can be moved out of the component, perhaps renamed to `depths` then where it is used as a prop, that named `depth`.

src/components/Token/ParallaxToken/ParallaxUseCasesElement.tsx
```diff
@@ -0,0 +1,44 @@
...
+    return generateParallaxWrapper(index, className)
+  })
+
+  return (
+    <div className={css.parallaxWrapper}>
+      <FrameImage className={css.baseImage} />
+      {parallaxWrappers}
+    </div>
+  )
+}
```
iamacook, line 42: This is hard to follow with the nested functions. Is there a reason for using them? I would suggest rendering them "normally", e.g.

```suggestion
const depths = [0, 1, 2] as const

const ParallaxGovernanceProcessElement = ({ items }: { items: Entry<TypeBaseBlockSkeleton, undefined, string>[] }) => {
  return (
    <div className={css.parallaxWrapper}>
      <FrameImage className={css.baseImage} />
      {items.map((_, index) => {
        const className = blockClassNames[index] || ''
        const depth = depths[index % 2]
        
        return (
          <ParallaxWrapper translateX={0} translateY={0} depth={depth} direction={-1} key={index}>
            <div className={`${css.movingElement} ${css[className]}`}>
              <Typography>{items[index].fields.caption}</Typography>
              {trailingShadow}
            </div>
          </ParallaxWrapper>
        )
      })}
    </div>
  )
}
```

src/components/Token/PulsingLogo/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+        alignItems: 'center',
+      }}
+    >
+      <motion.div
+        style={{
+          zIndex: 20,
+          width: `${circleDiameter}px`,
+          height: `${circleDiameter}px`,
+          borderRadius: '50%',
+          backgroundColor: '#1E3B2B',
```
iamacook, line 62: Can you not use a palette var. here?

src/components/Token/PulsingLogo/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+          duration: 2,
+          ease: 'easeInOut',
+        }}
+        src={`/images/Token/Hero/Logos/${imgUrl}`}
+        width={35}
+        style={{
+          zIndex: 30,
+        }}
+      />
+    </motion.div>
```
iamacook, line 99: This is very hard to follow. Can you break this down and add comments?

Some of the values are reused and the conditions should be abstracted to variables.

src/components/common/LinkButton/index.tsx
```diff
@@ -18,9 +18,11 @@ const LinkButton = ({
...
     >
       {children}
-      <div className={css.arrow}>
-        <AngleIcon />
-      </div>
+      {children ? (
+        <div className={css.arrow}>
+          <AngleIcon />
+        </div>
+      ) : undefined}
```
iamacook, line 25: Nit:
```suggestion
      ) : null}
```

src/components/common/RichText/index.tsx
```diff
@@ -32,18 +32,29 @@ const options: Options = {
...
+      })
+
+      return <Typography variant="h2">{content}</Typography>
     },
     [BLOCKS.HEADING_3]: (node: Heading3) => {
-      const text = node.content.find(isText)?.value
+      const content = node.content.filter(isText).map((node, index) => {
+        const isBold = node.marks.some((mark) => mark.type === 'bold')
+        return isBold ? <b key={index}>{node.value}</b> : <span key={index}>{node.value}</span>
+      })
```
iamacook, line 46: This logic is duplicated. Let's move it to a helper function.

src/components/common/RichText/index.tsx
```diff
@@ -32,18 +32,29 @@ const options: Options = {
...
 
       return (
-        <Typography id={kebabCase(text)} variant="h3" mt="60px">
-          {text}
+        <Typography variant="h3" id={kebabCase(node.content.find(isText)?.value)}>
```
iamacook, line 49: Is an `id` needed?

src/hooks/usePrefersReducedMotion.ts
```diff
@@ -0,0 +1,25 @@
+import { useState, useEffect } from 'react'
+
+const REDUCED_MOTION_MEDIA_QUERY = '(prefers-reduced-motion: reduce)'
+
+const usePrefersReducedMotion = (): boolean => {
+  const [prefersReducedMotion, setPrefersReducedMotion] = useState<boolean>(false)
+
+  useEffect(() => {
+    const mediaQueryList = window.matchMedia(REDUCED_MOTION_MEDIA_QUERY)
```
iamacook, line 9: I think we need add a `window` definition check here, or at least make it optionally chained.

src/pages/[slug].tsx
```diff
@@ -0,0 +1,49 @@
...
+import { type LandingPageEntry } from '@/config/types'
+
+const LandingPage = (props: { landingPage: LandingPageEntry }) => {
+  if (!props.landingPage) return null
+
+  return <Page {...props} />
+}
+
+export const getStaticProps: GetStaticProps = async ({ params }) => {
+  const slug = params?.slug as string
```
Comment (iamacook) [inline src/pages/[slug].tsx:14]: `slug` can be `undefined` and this is not being handled. It would be better to do something like:

```suggestion
  const slugParam = params?.slug
  
  if (!slugParam) {
    return {
      notFound: true
    }
  }
  
  const slug = Array.isArray(slugParam) ? slugParam[0] : slugParam
```

I can see that this is done elsewhere in the project so I will leave it up to you whether to change this.

iamacook (commented):
Everything is looking good. I've left a few comments, some of which being nits that I'll leave to your discretion.

A lot of the `framer-motion` code is quite hard to follow. It would be beneficial to break some of it down and add comments. Revisiting it in the future will be quite difficult otherwise.

As a general note: it's hard to review such large diffs. If you can "atomicise" future PRs, it'd make things easier.

Nice work so far!

src/components/Token/Hero/token-hero-rings.json
DiogoSoaress: True. Probably Malay didn't have the formatter running on save actions.

src/components/Token/ParallaxToken/ParallaxUseCasesElement.tsx
DiogoSoaress: 100% thanks for the comment.
For some reason I thought I was making it more readable but I reckon that the nested functions are making it way less readable.

src/components/Token/PulsingLogo/index.tsx
DiogoSoaress: Not in this one. We don't have this value in the color palette.

src/components/common/RichText/index.tsx
DiogoSoaress: Yes. That becomes the anchor that the Table of Contents navigate to in the Blog.

src/pages/[slug].tsx
Comment (DiogoSoaress) [reply src/pages/[slug].tsx:14]: I remember we've discussed this in some other PR. I feel here the check is unnecessary. Reason being this code runs at build time and if there is no `slug` the page is not built and the error will be visible.

src/components/Token/Floaters/index.tsx
DiogoSoaress: Unnecessary. I've overlooked it when reviewing the animation code.

src/components/Token/Floaters/index.tsx
DiogoSoaress: Added some useful comments

iamacook (approved):
Thanks for clarifying the code! Just a couple small remarks but otherwise it all looks good!

src/components/Token/Floaters/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const requestRef = useRef<number>()
+
+  // Use a ref to store the last timestamp when the animation frame was called
+  const lastFrameTime = useRef<number>(0)
+
+  // TODO: move to a separate hook
+  const animate = useCallback(
+    (time: number) => {
+      // If it's the first frame, initialize lastFrameTime to the current time
+      if (!lastFrameTime.current) lastFrameTime.current = time
```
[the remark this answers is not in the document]
DiogoSoaress: Exactly. Only at first render.

PR #301, opened 2024-04-10
feat: update typography as per UI Kit

DiogoSoaress:
Update theme's typography as per latest UI Kit guide -> https://www.figma.com/file/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?type=design&node-id=10829-708&mode=design&t=zuBY8bCFKL1ILz9H-0

Typography feedback was collected here: https://www.notion.so/safe-global/Typography-Updates-QA-54b63bdeab744e80a5605402a85e7753

I've addressed in https://github.com/safe-global/safe-homepage/pull/368 the feedback I found unrelated with this change


DiogoSoaress:
@TanyaEfremova can you review the typography changes across the pages? The font change is very noticeable so I'd like to have your 👍 before merging it into production. Link -> https://update_typography--homepage.review.5afe.dev/

TanyaEfremova:
Some of the headings use font weight light (300), however, it was only intended for the body text.

DiogoSoaress:
> Some of the headings use font weight light (300), however, it was only intended for the body text.

You're right! Just fixed it. Can you check again?

PR #309, opened 2024-04-19
feat: SafeAtScale animation impl & bracket video assets

fernhook:
implemented expanding brackets, stats & ticker animation, switches to vertical orientation at < 600px breakpoint


fernhook:
recheck

PR #327, opened 2024-04-21
feat: Pushwoosh notification form

DiogoSoaress:
Implements a form that registers an email to receive Safe{Pass} notifications


src/components/Rewards/SignUpForm/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const SignUpForm = (props: BaseBlockEntry) => {
+  const { title, text } = props.fields
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault()
+
+    const target = e.target as typeof e.target & {
+      email: { value: string }
+    }
```
iamacook, line 17: Nit: it would be a bit cleaner to use `FormData` here if you want to use an uncontrolled component (and extract the name to a constant), e.g.

```ts
const data = new FormData(event.target);
const email = data.get(FIELD_NAME);
```

src/components/Rewards/SignUpForm/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    fetch(PUSHWOOSH_ENDPOINT, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        email: target.email.value,
+        application: process.env.NEXT_PUBLIC_PUSHWOOSH_APPLICATION_CODE,
```
iamacook, line 26: We should tag these emails so we know where they come from.

src/components/Rewards/SignUpForm/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      }),
+    })
+      .then((response) => {
+        if (response.ok) {
+          console.log('Email registered successfully.')
+        } else {
+          console.error('Failed to register email.')
+        }
```
iamacook, line 34: Do we want logs on prod.?

src/components/Rewards/SignUpForm/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      })
+      .catch((error) => {
+        console.error('Error:', error)
```
iamacook, line 37: Maybe we should give use feedback here?

src/components/Rewards/SignUpForm/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const { title, text } = props.fields
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault()
+
+    const target = e.target as typeof e.target & {
+      email: { value: string }
+    }
+
+    fetch(PUSHWOOSH_ENDPOINT, {
```
iamacook, line 19: I would suggest we move this out of the component.

iamacook (approved):

PR #339, opened 2024-04-29
feat: navigation bar v2

DiogoSoaress:
## What it solves
Implements the redesigned navigation to accommodate the increased number of navigation items 

## Figma
https://www.figma.com/file/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?type=design&node-id=12520-146768&mode=design&t=QuPtc9HmPismYbq0-0


src/components/common/Footer/index.tsx
```diff
@@ -48,8 +46,6 @@ const ecosystemItems = [
   {
     label: 'Explore',
     href: AppRoutes.ecosystem,
-    target: '_blank',
-    rel: 'noreferrer',
```
iamacook, line 52: Why do we not want to open these in a new window? If necessary, I'd suggest doing this in a separate PR as it is unrelated to the header.

src/components/common/Header/index.tsx
```diff
@@ -1,77 +1,133 @@
...
+                          <Paper className={css.menu}>
+                            <Menu items={items ?? []} onItemClick={closeMobileNavigation} />
+                          </Paper>
+                        </Fade>
+                      )}
+                    </Popper>
+                  </ButtonBase>
+                )}
+              </li>
+            </>
```
iamacook, line 128: I'd suggest moving this to a new component. You then won't need to do any ref magic.

src/components/common/Header/index.tsx
```diff
@@ -1,77 +1,133 @@
...
+                    <AccordionDetails>
+                      <Menu items={items ?? []} onItemClick={closeMobileNavigation} />
+                    </AccordionDetails>
+                  </Accordion>
+                ) : (
+                  // Desktop button
+                  <ButtonBase
+                    className={clsx(css.navButton, css.hideInMobile, { [css.active]: subMenuOpen === category })}
+                    disableRipple
+                    id={`${category}-button`}
```
iamacook, line 99: Why is the dynamic `id` needed?

src/components/common/Header/index.tsx
```diff
@@ -1,77 +1,133 @@
...
+                    aria-haspopup
+                    aria-controls={`${category}-popper`}
+                    aria-expanded={isOpen}
```
iamacook, line 102: If we are adding aria- props here, we should also add them to the other elements.

src/components/common/Header/index.tsx
DiogoSoaress: No use for it in fact. Will remove 👍

src/components/common/Footer/index.tsx
DiogoSoaress: I sneaked these changes in this PR because I was revisiting the links. These two pages should be internal routing as they pages from the website and not open in a new page as for example the help center or the grants page.
I will validate this approach

iamacook: You can leave it is for now but I was confused to see footer-related code here.

src/components/common/Header/index.tsx
DiogoSoaress: Extracted and removed the refs array

src/components/common/Header/index.tsx
DiogoSoaress: I only found missing the `aria-label` attribute. Any specific attribute missing that you can think of?

PR #341, opened 2024-05-13
feat: add Safe{Pass} button to the navigation

DiogoSoaress:
## Figma
https://www.figma.com/file/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?type=design&node-id=12520-146768&mode=design&t=SrG7ZuhjV8XVWZQK-0

### DIFF
--- a/src/components/common/Header/NavigationButton.tsx
+++ b/src/components/common/Header/NavigationButton.tsx
@@ -17,20 +17,22 @@ const NavigationButton = ({ category, items, subMenuOpen, onItemClick }: Navigat
 
   const isOpen = subMenuOpen === category
 
+  const categoryString = typeof category === 'string' ? category : 'unknown category'
+
   return (
     <ButtonBase
       className={clsx(css.navButton, css.hideInMobile, { [css.active]: isOpen })}
       disableRipple
-      aria-label={category}
+      aria-label={categoryString}
       aria-haspopup
-      aria-controls={`${category}-popper`}
+      aria-controls={`${categoryString}-popper`}
       aria-expanded={isOpen}
       ref={buttonRef}
     >
       <div className={css.navButton}>{category}</div>
       <Popper
         className={css.hideInMobile}
-        id={`${category}-popper`}
+        id={`${categoryString}-popper`}
         open={isOpen}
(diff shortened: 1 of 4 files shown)


TanyaEfremova:
All good!

src/components/common/Header/NavigationButton.tsx
```diff
@@ -17,20 +17,22 @@ const NavigationButton = ({ category, items, subMenuOpen, onItemClick }: Navigat
 
   const isOpen = subMenuOpen === category
 
+  const categoryString = typeof category === 'string' ? category : 'unknown category'
```
iamacook, line 20: All `catgory`s are capitalised words. Do we wants to use `"Unknown"` instead?

src/components/common/Header/index.tsx
```diff
@@ -46,40 +46,38 @@ const Header = () => {
...
-                    items={items ?? []}
-                    subMenuOpen={subMenuOpen}
-                    onItemClick={closeMobileNavigation}
-                  />
-                )}
-              </li>
-            </>
+          {navCategories.map(({ category, items, href }) => (
+            <li
+              key={typeof category === 'string' ? category : href}
```
iamacook, line 51: Nit: I'd suggest keeping the key simple:

```suggestion
              key={href}
```

src/components/common/Header/index.tsx
```diff
@@ -46,40 +46,38 @@ const Header = () => {
...
+              onMouseEnter={toggleCategoryOpen(typeof category === 'string' ? category : null)}
+              onFocus={toggleCategoryOpen(typeof category === 'string' ? category : null)}
+              onMouseLeave={toggleCategoryOpen(null)}
+              onBlur={toggleCategoryOpen(null)}
```
iamacook, line 55: You can abstract these to two variables before rendering, e.g.:

```ts
const onOpen = toggleCategoryOpen(typeof category === 'string' ? category : null)
const onClose = toggleCategoryOpen(null)
```

DiogoSoaress: I've undone these changes. Included `JSX.Element` as a possible input type instead and added an early return for it.

PR #353, opened 2024-05-27
feat: live updates on blog posts and blog landing page

DiogoSoaress:
## What is changed

Enable live updates on blog & press release posts and on the blog landing page
- Fetch CMS API when rendering a blog post
- Fetch CMS API when rendering the blog landing page

A new `isDraft` boolean was added to the BlogPost content type on the CMS to manage live what content should be visible/hidden in production.


src/components/Blog/BlogHome/index.tsx
```diff
@@ -1,26 +1,60 @@
...
+    client
+      .getEntries<TypePostSkeleton>({
+        content_type: 'post',
+        order: ['-fields.date'],
+      })
+      .then((entry) => {
+        setLocalAllPosts(entry.items)
+      })
+      .catch(console.error)
+  }, [])
```
iamacook, line 55: Nit: I would move these to respective hooks. They are data fetching-relative so it would also be beneficial to have some form of loading/error flags. Using an external library such as SWT or React Query might be the way to go.

src/components/Blog/Post/index.tsx
```diff
@@ -19,11 +19,27 @@ import { type Document as ContentfulDocument } from '@contentful/rich-text-types
...
+  useEffect(() => {
+    client
+      .getEntry(blogPost.sys.id)
+      .then((entry) => {
+        if (isEntryTypePost(entry)) {
+          setPost(entry)
+        }
+      })
+      .catch(console.error)
+  }, [blogPost.sys.id])
```
iamacook, line 42: This could leverage the suggested hook.

DiogoSoaress: moved to `useBlogPost`

src/components/Blog/BlogHome/index.tsx
DiogoSoaress: moved both `useEffect` to hooks

src/components/Blog/BlogHome/index.tsx
```diff
@@ -1,26 +1,36 @@
...
-  featuredPost: BlogPostEntry
-  mostPopular: BlogPostEntry[]
-  allPosts: BlogPostEntry[]
+  blogHome: BlogHomeEntry
+  allPosts: EntryCollection<TypePostSkeleton, undefined, string>
 }
 
-const BlogHome = (props: BlogHomeProps) => {
-  const { featuredPost, mostPopular, allPosts, metaTags } = props
+const isDraft = (post: BlogPostEntry) => post.fields.isDraft
```
iamacook, line 25: Nit: I would move this to the relevant helper file.

src/hooks/useAllPosts.ts
```diff
@@ -0,0 +1,23 @@
+import { type TypePostSkeleton } from '@/contentful/types'
+import client from '@/lib/contentful'
+import { type EntryCollection } from 'contentful'
+import { useEffect, useState } from 'react'
+
+// FIXME: This hook should make use of useSWR instead of useState
```
iamacook, line 6: Nit: after discussion, I understand this comment but I'd suggest we give a brief description of why it is currently using `useState`.

iamacook (approved):

src/hooks/useAllPosts.ts
DiogoSoaress: I explain it [here](https://github.com/safe-global/safe-homepage/pull/353#issuecomment-2137523799)

PR #359, opened 2024-05-30
feat: Pressroom live updates

DiogoSoaress:
Allow to live update the Pressroom content from the CMS


src/hooks/usePressRoom.ts
```diff
@@ -0,0 +1,8 @@
+import useSWR from 'swr'
+import client from '@/lib/contentful'
+import type { TypePressRoomSkeleton } from '@/contentful/types'
+import type { PressRoomEntry } from '@/components/Pressroom'
+
+const pressRoomFetcher = (id: string) => client.getEntry<TypePressRoomSkeleton>(id)
+
+export const usePressRoom = (id: string, fallbackData: PressRoomEntry) => useSWR(id, pressRoomFetcher, { fallbackData })
```
iamacook, line 8: This only differs to [`useBlogHome`](https://github.com/safe-global/safe-homepage/blob/main/src/hooks/useBlogHome.ts) by type. We could use the same code and pass a generic, something like `useFetchEntry<T>`. What do you think?

src/components/Pressroom/PressReleases/index.tsx
```diff
@@ -7,24 +6,31 @@ import ShowMoreButton from '@/components/common/ShowMoreButton'
...
+import { useAllPosts } from '@/hooks/useAllPosts'
+import type { EntryCollection } from 'contentful'
+import type { TypePostSkeleton } from '@/contentful/types'
 
 const categories = ['Safe{Core}', 'Safe{Wallet}', 'Safe{DAO}', 'Ecosystem', 'Institutional', 'Internal']
 
 const PAGE_LENGTH = 4
 
-const PressReleases = ({ allPosts }: { allPosts: BlogPostEntry[] }) => {
+const PressReleases = ({ allPosts }: { allPosts: EntryCollection<TypePostSkeleton, undefined, string> }) => {
```
iamacook, line 19: Nit: can we import this type from somewhere else, e.g. where the SWR hook for all posts is?

src/components/Pressroom/index.tsx
```diff
@@ -24,26 +22,28 @@ import {
...
-import { type Entry } from 'contentful'
+import type { EntryCollection, Entry } from 'contentful'
+import { usePressRoom } from '@/hooks/usePressRoom'
 
 export type PressRoomEntry = Entry<TypePressRoomSkeleton, undefined, string>
 
 export type PressRoomProps = {
   pressRoom: PressRoomEntry
-  allPosts: BlogPostEntry[]
+  allPosts: EntryCollection<TypePostSkeleton, undefined, string>
```
iamacook, line 32: Nite: as above regarding type import.

src/components/Pressroom/PressReleases/index.tsx
DiogoSoaress: Good catch ! It was about time to create that type.

src/hooks/usePressRoom.ts
DiogoSoaress: Done. I've called it `useLandingPageContent`

src/components/Pressroom/PressReleases/index.tsx
```diff
@@ -7,24 +6,30 @@ import ShowMoreButton from '@/components/common/ShowMoreButton'
...
 
+  const { localAllPosts } = useAllPosts(allPosts)
+
   const filteredPosts = useMemo(() => {
-    return !selectedTag ? allPosts : allPosts.filter((post) => containsTag(post.fields.tags, selectedTag))
-  }, [allPosts, selectedTag])
+    const pressPosts = localAllPosts.items.filter(isPressReleasePost)
+
+    return !selectedTag ? pressPosts : pressPosts.filter((post) => containsTag(post.fields.tags, selectedTag))
+  }, [localAllPosts, selectedTag])
```
iamacook, line 29:
```suggestion
  }, [localAllPosts.items, selectedTag])
```

src/hooks/useLandingPageContent.ts
```diff
@@ -0,0 +1,24 @@
+import useSWR from 'swr'
+import client from '@/lib/contentful'
+import type { Entry, EntrySkeletonType } from 'contentful'
+
+const fetcher = <T extends EntrySkeletonType>(id: string) => client.getEntry<T>(id, { include: 3 })
```
iamacook, line 5: Nit: I would also allow customisation of the `includes` value via props. However, you can default it to `3`.

src/hooks/useLandingPageContent.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @template T - The type of the Contentful entry. It must extend EntrySkeletonType.
+ * @template U - The type of the fallback data. It must be an instance of Entry with T as the first type parameter.
+ *
+ * @param {string} id - The ID of the Contentful entry to fetch.
+ * @param {U} fallbackData - The fallback data to use while the Contentful entry is being fetched.
+ *
+ * @returns {useSWR.Response} - The response from useSWR, which includes the data and error (if any).
+ */
+export const useLandingPageContent = <T extends EntrySkeletonType, U extends Entry<T, undefined, string>>(
```
iamacook, line 18: Nit: this name is a bit generic. Maybe something like `useClientEntry` or something like that?

PR #363, opened 2024-06-06
fix: Header breakpoint does not give room to all buttons

DiogoSoaress:
## What it solves
- Increases the mobile responsive breakpoint in the Header component.
- Semantically split the HTML and CSS for the Header `<nav>` and `<ul>`

## Screenshots
**Before**
Screenshot 2024-06-04 at 11 25 48


src/hooks/useIsSmallScreen.ts
```diff
@@ -0,0 +1,4 @@
+import { useMediaQuery } from '@mui/material'
+
+export const useIsExtraSmallScreen = () => useMediaQuery('(max-width:600px)')
+export const useIsSmallScreen = () => useMediaQuery('(max-width:900px)')
```
iamacook, line 4: Nit: I would keep using the theme as the source of truth for breakpoints (I know that in CSS files we have to use the pixel values though), e.g.

```suggestion
export const useIsSmallScreen = () => useMediaQuery((theme: Theme) => theme.breakpoints.down('sm'))
export const useIsMediumScreen = () => useMediaQuery((theme: Theme) => theme.breakpoints.down('md'))
```

Note: "extra small" is [defined as `0px` by MUI](https://mui.com/material-ui/customization/breakpoints/#default-breakpoints) so I adjusted the names.

We should also generify this file name.

DiogoSoaress: Fixed.

Should I move each hook to a separate file? I am wondering if it still makes sense to start the filename by `use...`

iamacook: > Should I move each hook to a separate file?

I'll leave it up to your discretion. Considering the repetition, you could even extract the "base" logic to a hook (which you name the file) and use it/derive the others from it, e.g.

```ts
// useMaxWidth.ts
import { useMediaQuery } from '@mui/material'
import type { Breakpoint, Theme } from '@mui/material'

export const useMaxWidth = (key: number | Breakpoint) => useMediaQuery((theme: Theme) => theme.breakpoints.down(key))

// If you want to derive the hooks instead of using the above directly
export const useIsSmallScreen = () => useMaxWidth('sm')
export const useIsMediumScreen = () => useMaxWidth('md')
```

iamacook (approved):

TanyaEfremova:
LFG. Looks good!

Issue #380, opened 2024-06-26
[Dataroom] Transactions of chain


DiogoSoaress:
Part of #354 

### Figma
https://www.figma.com/design/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?node-id=13958-10672&t=1qAGH4DQITJXjbQ5-4

### Animation description
Numbers slide up replacing the previous values

### Screenshot
<img width="435" alt="Screenshot 2024-06-26 at 11 53 52" src="https://github.com/safe-global/safe-homepage/assets/32431609/51637d04-1e10-47a4-b7be-cb7075552fea">

_Approximate estimation: 1/2 day_

PR #388, opened 2024-06-27
Add Transaction On Chain Component + Auxiliary Components

Malayvasa:
Related to #380


DiogoSoaress (commented):
Some comments before a thorough review:
- Can you please use CSS modules to apply styles instead of MUI component attributes whenever possible?
- Can you write the React components as a const instead of a function? Please look at any other component to see the style that is used across the codebase.
- Constants should follow the name convention: capital letters with underscore separators. Example: `const FOO_BAR = 42`

DiogoSoaress:
@Malayvasa In that case you can generate the style object from a function that receives `direction` as a parameter

Issue #379, opened 2024-06-26
[Dataroom] Volume transferred


DiogoSoaress:
Part of #354 

### Figma
https://www.figma.com/design/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?node-id=13958-10672&t=1qAGH4DQITJXjbQ5-4

### Animation description
Numbers slide in left and right ends of the screen

### Screenshot
<img width="436" alt="Screenshot 2024-06-26 at 11 52 13" src="https://github.com/safe-global/safe-homepage/assets/32431609/68e12ef9-0401-42f7-8c93-cac19372aaee">

_Approximate estimation: done_

PR #400, opened 2024-07-03
Add mobile support for volume transferred component

Malayvasa:
Related to #379 

_Note : Please disable TransactionsOnChain component while previewing, as it causes width to increase more than 100vh and displaces all other sections._


DiogoSoaress (commented):
Shouldn't we make the sliding numbers smaller in mobile? It's hard to read with this font size

Issue #376, opened 2024-06-26
[Dataroom] Hero


DiogoSoaress:
Part of #354

### Figma
https://www.figma.com/design/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?node-id=13958-10672&t=1qAGH4DQITJXjbQ5-4

### Animation description
Interactive 3D 

_Approximate estimation:_ 2 days

PR #405, opened 2024-07-04
Add Hero Section Component

Malayvasa:
Related to #376


src/components/DataRoom/Hero/SafeLogo.tsx
```diff
@@ -0,0 +1,45 @@
...
+import type { MotionValue } from 'framer-motion'
+
+type SafeLogoProps = {
+  yPosition: MotionValue<number>
+  isMobile: boolean
+  offset: number
+}
+
+export function SafeLogo({ isMobile, offset, yPosition }: SafeLogoProps) {
+  const { nodes } = useGLTF('/Safe_Logo.glb') as any
```
DiogoSoaress, line 15: Can we avoid casting to `any` ?

src/components/DataRoom/Hero/SafeLogo.tsx
```diff
@@ -0,0 +1,45 @@
...
+import type { Group, Mesh } from 'three'
+import type { MotionValue } from 'framer-motion'
+
+type SafeLogoProps = {
+  yPosition: MotionValue<number>
+  isMobile: boolean
+  offset: number
+}
+
+export function SafeLogo({ isMobile, offset, yPosition }: SafeLogoProps) {
```
DiogoSoaress, line 14: You don't need to pass `isMobile` as a prop. Can use the hook inside this component

src/components/DataRoom/Hero/index.tsx
```diff
@@ -1,13 +1,43 @@
...
-  </Container>
-)
+const Hero = ({ title, text }: BaseBlock) => {
+  const containerRef = useRef<HTMLDivElement>(null)
+  const { scrollYProgress } = useScroll({
+    target: containerRef,
+    offset: ['start end', 'end start'],
+  })
+
+  const isMobile = useMaxWidth(900)
```
DiogoSoaress, line 16: Use https://github.com/safe-global/safe-homepage/blob/b970a311bd8d219beb080772092d591d0dfcb3d0/src/hooks/useMaxWidth.ts#L7

src/components/DataRoom/Hero/styles.module.css
```diff
@@ -0,0 +1,71 @@
...
+    height: calc(100dvh - 95px);
+  }
+}
+
+.content {
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  gap: 2rem;
+  margin-bottom: 3rem;
```
DiogoSoaress, line 56: Can we not use rem units?

src/components/DataRoom/Hero/utils.ts
DiogoSoaress, line 1: Can you create a /utils folder separate the different utils into different files inside of it?

src/components/DataRoom/Hero/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+        delta,
+      )
+    }
+
+    state.camera.position.set(state.camera.position.x, state.camera.position.y, 5)
+    state.camera.lookAt(0, 0, 0)
+  })
+
+  return null
+}
```
DiogoSoaress, line 51: Is this a component? Please add a description to these utils and add return types to the functions

src/components/DataRoom/Hero/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    // Apply damping to camera position based on mouse movement on desktop
+    if (!isMobile) {
+      easing.damp3(
+        state.camera.position,
+        [1 - mouse.current.x * 2, 1 - mouse.current.y * 2, 20 + Math.atan(mouse.current.y * 4)],
+        0.3,
+        delta,
+      )
+    }
```
DiogoSoaress, line 44: Why do we need this? I would cross check with Tanya if it's really necessary.

src/components/DataRoom/Hero/AsciiScene.tsx
```diff
@@ -0,0 +1,46 @@
...
+  yPosition,
+  zoomLevel,
+}: {
+  isMobile: boolean
+  yPosition: MotionValue<number>
+  zoomLevel: MotionValue<number>
+}) => {
+  return (
+    <Canvas orthographic camera={{ position: [0, 0, 10], zoom: isMobile ? 70 : 80 }}>
+      <CameraController zoomLevel={zoomLevel} isMobile={isMobile} />
```
DiogoSoaress, line 19: Can you provide some descriptions in the PR or in the components? It's hard to follow otherwise

DiogoSoaress (changes requested):

src/components/DataRoom/Hero/AsciiWrapper.tsx
```diff
@@ -0,0 +1,12 @@
+import { useEffect, useMemo } from 'react'
+import { ASCIIEffect } from './utils'
+
+export default function AsciiWrapper() {
```
DiogoSoaress, line 4:
```suggestion
const AsciiWrapper = () =>  {
```
and add the return type

src/components/DataRoom/Hero/utils/cameraController.ts
DiogoSoaress, line 1: This should not be a component as it does not return any jsx but rather a custom hook (`useCameraController`)

src/components/DataRoom/Hero/utils/easing.ts
```diff
@@ -0,0 +1,44 @@
+import { Vector3 } from 'three'
+
+export function damp(
+  current: Vector3,
+  target: number | [x: number, y: number, z: number] | Vector3,
+  smoothTime: number = 0.3,
+  delta: number = 1 / 60,
+  maxSpeed: number = Infinity,
+  easing: (t: number) => number = (t) => t,
+  eps: number = 0.001,
```
DiogoSoaress, line 10: I suggest reducing the number of parameters this function accepts (only `current`, `target` and `delta`) and define the default parameter values as constants in this file.

```
const SMOOTH_TIME_DEFAULT = 0.3
const DELTA_DEFAULT = 1 / 60
const MAX_SPEED_DEFAULT = Infinity
const EPS_DEFAULT = 0.001
```

The easing function can be defined in this file but as a separate function.

DiogoSoaress (changes requested):
The code looks good and the animation works smoothly 👏 
Just a couple of _easy-to-address_ refactor comments

src/components/DataRoom/Hero/utils/cameraController.ts
DiogoSoaress: Yep, I got the same. Let's leave it as it is

DiogoSoaress (approved):
LGTM

PR #409, opened 2024-07-07
style(gas-station): design fixes

DiogoSoaress:
Addresses the following style fixes:
- Center text block in Hero section
- Smooth the edges of the video + slider
- Have the Slider occupy the container full width
- Between sections spacing


iamacook (commented):
Everything looks good but two things I noticed on the deployment:

The video and carousel "end" abruptly. Can they transition into the background instead?

image

---

The doc-related cards stack. I think they should either span the entire width of the container, or be positioned horizontally:

image

iamacook:
> I've allowed the background blobs to overlap the slider giving it a really nice and smooth effect. Also added a gradient to the right side of the video.

There is still the issue that the video ends "abruptly" and carousel doesn't span the entire width. The designs depict them spanning the entire page. cc @TanyaEfremova

image

image

DiogoSoaress:
The page content is limited to `max-width: 1630px;` because of https://github.com/safe-global/safe-homepage/blob/c2a52994205f79c5acf196277dc12ee5d4dab201/src/components/common/PageLayout/styles.module.css#L2 which impacts the video and the carrousel, so the video will need to be cut in larger viewports eventually.

What could be a elegant way to handle it?

iamacook:
> What could be a elegant way to handle it?

You could blur to both sides of the video/carousel but maybe @TanyaEfremova has a better idea.

DiogoSoaress:
I've added gradients on both sides and I think it looks much better now! Thanks for making a point. I will ping Tanya to give her opinion

iamacook (approved):

TanyaEfremova:
I like the suggestion, and I think, we should proceed with that, however, on the link above I cannot see the gradient within the ribbons.
Possible to make something [like this](https://www.figma.com/design/WNczwX5oGaGi01Q8EvJKm6/Gas-Station?node-id=12517-702&t=QjtXrAPaslME4imL-1)? I replicated it in design. If we can use a different color for the gradient, like in the video itself `#030D02`, it would look even better and more "natural".

DiogoSoaress:
Can you include this blob in the design? I think that's what emphasized so much the edges of the video + slider container. I suggest we move it to a new issue as an improvement

Screenshot 2024-07-08 at 13 06 29

PR #411, opened 2024-07-07
feat(dataroom): Fetch stats from Dune

DiogoSoaress:
Leverages the Dune query created by the Data team to pull the real time values.

Numbers are passed by the `SafeDataRoomContext` context and can be pulled from the `useSafeDataRoomStats` custom hook


src/components/DataRoom/CryptoPunks/utils.ts
```diff
@@ -0,0 +1,9 @@
+export function getColor() {
+  const chance = Math.random()
+
+  if (chance <= 0.24) return '#12FF80' // Green (14% CHANCE)
+  if (chance <= 0.17) return '#12A154' // Dark Green (3% CHANCE)
+  if (chance <= 0.2) return '#124228' // Darker Green (3% CHANCE)
+
+  return 'currentColor'
+}
```
iamacook, line 9: Is the base branch correct? I remember a similar function from another PR.

src/components/DataRoom/TransactionsOnChain/index.tsx
```diff
@@ -7,16 +7,21 @@ import css from './styles.module.css'
...
 
-// Will be replaced with the actual value in a future PR
-const TRANSACTIONS_AMOUNT = 1.75
+const TRANSACTIONS_ON_CHAIN_PERCENTAGE_FALLBACK = 1.75
 
 const TransactionsOnChain = ({ text, link }: BaseBlock) => {
+  const { onChainTransactionsPercentage } = useSafeDataRoomStats()
+
+  const value = onChainTransactionsPercentage || TRANSACTIONS_ON_CHAIN_PERCENTAGE_FALLBACK
+  const displayValue = Number(toPercentage(value))
```
iamacook, line 19: This confused me seeing you convert a percentage to a number. If you choose not to adjust the name of `toPercentage` (see my other comment), I'd suggest doing the maths here for more clarity.

src/contexts/SafeDataRoomContext.ts
```diff
@@ -0,0 +1,19 @@
...
+  onChainTransactionsPercentage: number | null
+}
+
+const SafeDataRoomContext = createContext<SafeDataRoomStats>({
+  tvlToGDPPercentage: null,
+  usdcPercentageStored: null,
+  cryptoPunksStoredPercentage: null,
+  totalVolumeTransfered: null,
+  onChainTransactionsPercentage: null,
+})
```
iamacook, line 17: I understang that you wanted to localise the constants but maybe the fallback values are better located here. You then don't need to repeat the `value | fallback` logic in every file. What do you think?

src/hooks/useSafeDataRoomStats.ts
```diff
@@ -0,0 +1,37 @@
...
+  tvl_perc_world_gdp: number
+  usdc_in_safes: number
+}
+
+const QUERY_ID_DATAROOM_STATS = 3864414
+
+export const fetchDataRoomStats = async (): Promise<DuneDataRoomStats | null> => {
+  return fetch(duneQueryUrlBuilder(QUERY_ID_DATAROOM_STATS, DUNE_API_KEY))
+    .then((res) => res.json())
+    .then((data) => data.result.rows[0])
```
iamacook, line 31: You aren't handling response errors here. I'd also suggest including types - ideally validating the response because any changes to Dune will cause an error.

```suggestion
  return fetch(duneQueryUrlBuilder(QUERY_ID_DATAROOM_STATS, DUNE_API_KEY))
    .then((res) => {
      if (res.ok) {
        return res.json()
      } else {
        throw new Error('Response error')
      })
    // Validate response instead of casting
    .then((data) => data.result.rows[0] as DuneDataRoomStats)
```

src/lib/formatNumberWithThousandSeparators.ts
```diff
@@ -0,0 +1,14 @@
...
+ * This function takes a number, discards its decimal part, and then formats it
+ * according to the 'en-US' locale, which adds thousand separators to the integer part of the number.
+ *
+ * @param {number} num - The number to format.
+ * @returns {string} The formatted number with thousand separators.
+ */
+export const formatNumberWithThousandSeparators = (num: number): string => {
+  const integerPart = Math.floor(num) // Discard the decimal part
+
+  return new Intl.NumberFormat('en-US').format(integerPart)
```
iamacook, line 13: Note: if the locale is `undefined`, this will use that of the browser. I assume you always want to format to `en-US` though.

src/lib/toPercentage.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Converts a decimal number to a percentage string.
+ *
+ * @param {number} num - The decimal number to convert.
+ * @param {number} [decimalPlaces=2] - Optional. The number of decimal places for the percentage string.
+ * @returns {string} The converted number as a percentage string with the specified number of decimal places.
+ */
+export const toPercentage = (num: number, decimalPlaces: number = 2): string => {
+  return (num * 100).toFixed(decimalPlaces)
+}
```
iamacook, line 10: The usage of this initially confused me. I undestang that it requires a float after reading the comment but it's only multiplies by 100. I'd suggest being a bit more declartive with the name, e.g. `decimalToPercentage` or also allow integers.

src/components/DataRoom/USDCStorage/index.tsx
```diff
@@ -33,6 +36,9 @@ const USDCStorage = ({ title, text, link, image }: BaseBlock) => {
   const xTransformCanvas = useTransform(mapYProgress, [0.8, 1], getTransformParams(['0%', '100%']))
   const opacity = useTransform(scrollYProgress, [0.75, 1], [1, 0])
 
+  const value = usdcPercentageStored || USDC_PERCENTAGE_STORED_FALLBACK
+  const displayValue = `${toPercentage(value, 1)}%`
```
iamacook, line 40: I somewhat expected `toPercentage` to return a string with "%" but I'll leave it up to you if it should or not.

src/components/DataRoom/VolumeTransferred/index.tsx
```diff
@@ -20,6 +23,9 @@ const VolumeTransferred = ({ title, text, link }: BaseBlock) => {
   const opacityLTR = useTransform(scrollYProgress, [0.25, 0.3, 0.7, 0.75], [0, 1, 1, 0])
   const opacityRTL = useTransform(scrollYProgress, [0.25, 0.3, 0.7, 0.75], [0, 1, 1, 0])
 
+  const volumeTransferred = totalVolumeTransfered || VOLUME_TRANSFERRED_FALLBACK
+  const displayValue = `$${formatNumberWithThousandSeparators(volumeTransferred)}`
```
iamacook, line 27: Note: `Intl.NumberFormat` can return format currency according to locale. Providing `volumeTransferred` only ever has 2 decimal places and you don't change the locale of `formatNumberWithThousandSeparators` from `en-US`, this should be correct. However, you might want to consider a currency formatter to be safe.

src/components/DataRoom/CryptoPunks/utils.ts
DiogoSoaress: ahh!! Leftover from the rebase.

src/contexts/SafeDataRoomContext.ts
DiogoSoaress: I am not convinced the context would be the right place to have the fallbacks. What about in `src/pages/dataroom.tsx` ? Or I could pass the fallback value in the content like done for the Homepage https://github.com/safe-global/safe-homepage/blob/3b5deb0adf30ab3d717c03a68cc00084b2111887/src/content/home.json#L33

src/lib/formatNumberWithThousandSeparators.ts
DiogoSoaress: Yes, I think we want to be consistent with that `en-US` formatting. I will confirm with the team just to be sure

DiogoSoaress: It is also applied in this other formatter https://github.com/safe-global/safe-homepage/blob/3b5deb0adf30ab3d717c03a68cc00084b2111887/src/lib/formatValue.ts#L8 because one day the numbers were being displayed in a weird format. See https://5afe.slack.com/archives/C05LQG2MPSQ/p1693811373468229

src/lib/toPercentage.ts
DiogoSoaress: I will rename it as suggested because you're right: it is only intended to be used only for percentages. 

Question: do you think I need to validate it receives a decimal number?

src/components/DataRoom/USDCStorage/index.tsx
DiogoSoaress: I will pass the "%" as optional

src/components/DataRoom/TransactionsOnChain/index.tsx
```diff
@@ -7,15 +7,21 @@ import css from './styles.module.css'
...
-// Will be replaced with the actual value in a future PR
-const TRANSACTIONS_AMOUNT = 1.75
+const TRANSACTIONS_ON_CHAIN_PERCENTAGE_FALLBACK = 1.75
 
 const TransactionsOnChain = ({ text, link }: BaseBlock) => {
+  const { onChainTransactionsPercentage } = useSafeDataRoomStats()
+
+  const value = onChainTransactionsPercentage || TRANSACTIONS_ON_CHAIN_PERCENTAGE_FALLBACK
+  // Converts to percentage with 2 decimal place
+  const percentageValue = +(value * 100).toFixed(2)
```
iamacook, line 19:
```suggestion
  const percentageValue = (value * 100).toFixed(2)
```

src/contexts/SafeDataRoomContext.ts
iamacook: I think it's fine as is, I just noticed the pattern across files that could be centralised. Keep the fallbacks where used also makes sense.

src/hooks/useSafeDataRoomStats.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const fetchDataRoomStats = async (): Promise<DuneDataRoomStats | null> => {
+  return (
+    fetch(duneQueryUrlBuilder(QUERY_ID_DATAROOM_STATS, DUNE_API_KEY))
+      .then((res) => {
+        if (!res.ok) throw new Error('Response error')
+
+        return res.json()
+      })
+      // Validate response instead of casting
```
iamacook, line 36: I'd suggest either removing this (if you're happy with casting) or marking it as a TODO.

iamacook (approved):

src/components/DataRoom/TransactionsOnChain/index.tsx
DiogoSoaress: This casts the `string` to `number`

DiogoSoaress: `toFixed` returns a `string`

iamacook: I overlooked the prop type of `CounterContainer`.

src/hooks/useSafeDataRoomStats.ts
DiogoSoaress: 100% it was overlooked 😅

Issue #381, opened 2024-06-26
[Dataroom] Industry comparison


DiogoSoaress:
Part of #354 

### Figma
https://www.figma.com/design/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?node-id=13958-10672&t=1qAGH4DQITJXjbQ5-4

### Animation description
Green section slides from the right
Grid with dots highlight to follow the mouse movement

### Screenshot
<img width="441" alt="Screenshot 2024-06-26 at 11 55 03" src="https://github.com/safe-global/safe-homepage/assets/32431609/07175915-d96c-4f36-bedc-811eb1df0d45">

_Approximate estimation: 1 day_

PR #427, opened 2024-07-30
feat: 🎸 Add Industry Comparison Section

Malayvasa:
Related to #381


src/components/DataRoom/IndustryComparison/DotGrid.tsx
```diff
@@ -0,0 +1,107 @@
...
+      }
+    }
+
+    const lerpMousePosition = () => {
+      setLerpedMousePosition((prev) => ({
+        x: lerp(prev.x, mousePosition.x),
+        y: lerp(prev.y, mousePosition.y),
+      }))
+      requestAnimationFrame(lerpMousePosition)
+    }
```
iamacook, line 51: Due to the dependencies, I'd suggest we move memoise these in `useCallback`s.

src/components/DataRoom/IndustryComparison/DotGrid.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const dots = useMemo(() => createDots(dimensions, isMobile), [dimensions, isMobile])
+
+  useEffect(() => {
+    const updateDimensions = () => {
+      const newDimensions = updateCanvasDimensions(canvasRef, containerRef)
+      setDimensions(newDimensions)
+    }
+
+    const handleMouseMove = (e: MouseEvent) => {
+      if (canvasRef.current && containerRef.current) {
```
iamacook, line 31: Is `canvasRef.current` required for the mouse position? The value isn't used inside the handler.

src/components/DataRoom/IndustryComparison/DotGrid.tsx
```diff
@@ -0,0 +1,107 @@
...
+        const rect = containerRef.current.getBoundingClientRect()
+        setMousePosition({ x: e.clientX - rect.left, y: e.clientY - rect.top })
+      }
+    }
+
+    const updateMobileMousePosition = () => {
+      if (scrollYProgress && isMobile && containerRef.current) {
```
iamacook, line 38: We could use early returns here to reduce indentation, e.g.
```suggestion
      if (canvasRef.current && containerRef.current) return
      const rect = containerRef.current.getBoundingClientRect()
      setMousePosition({ x: e.clientX - rect.left, y: e.clientY - rect.top })
    }

    const updateMobileMousePosition = () => {
      if (scrollYProgress && isMobile && containerRef.current) return
```

src/components/DataRoom/IndustryComparison/DotGrid.tsx
```diff
@@ -0,0 +1,107 @@
...
+        setMousePosition({ x: dimensions.width - dimensions.width / 8, y: progress * height })
+      }
+    }
+
+    const lerpMousePosition = () => {
+      setLerpedMousePosition((prev) => ({
+        x: lerp(prev.x, mousePosition.x),
+        y: lerp(prev.y, mousePosition.y),
+      }))
+      requestAnimationFrame(lerpMousePosition)
```
iamacook, line 50: We need to clean this up with `cancelAnimationFrame` on unmount.

src/components/DataRoom/IndustryComparison/DotGrid.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    return () => {
+      window.removeEventListener('resize', updateDimensions)
+      if (isMobile && scrollYProgress) {
+        scrollYProgress.clearListeners()
+      } else {
+        container?.removeEventListener('mousemove', handleMouseMove)
+      }
+    }
+  }, [mousePosition.x, mousePosition.y, containerRef, isMobile, scrollYProgress, dimensions.width])
```
iamacook, line 71: There's a lot going on here in one `useEffect`. If it's possible, breaking it down would be ideal. Memoising the methods in `useCallback`s would help.

src/components/DataRoom/IndustryComparison/DotGrid.tsx
iamacook, line 51: Afaiu, this is related to drawing the dots? Could we locate them together?

src/components/DataRoom/IndustryComparison/utils/updateCanvasDimensions.ts
```diff
@@ -0,0 +1,20 @@
+export const updateCanvasDimensions = (
+  canvasRef: React.RefObject<HTMLCanvasElement>,
+  containerRef: React.RefObject<HTMLDivElement>,
```
iamacook, line 3: Instead of passing the ref, I think we can use the element directly, e.g.
```suggestion
  canvas: HTMLCanvasElement,
  container: HTMLDivElement,
```

src/components/DataRoom/IndustryComparison/utils/updateCanvasDimensions.ts
```diff
@@ -0,0 +1,20 @@
...
+    canvasRef.current.style.height = `${clientHeight}px`
+
+    const ctx = canvasRef.current.getContext('2d')
+    if (ctx) {
+      ctx.scale(pixelRatio, pixelRatio)
+    }
+    return { width: clientWidth, height: clientHeight }
+  }
+  return { width: 0, height: 0 }
+}
```
iamacook, line 20: I'd suggest we split this - it's updating the canvas, but also getting the dimensions of the container (used in the component). I'd suggest we have two functions: something like `getContainerDimensions` (and maybe we don't need to store that in state) and then a function to update the canvas.

DiogoSoaress (changes requested):

src/components/DataRoom/IndustryComparison/styles.module.css
```diff
@@ -0,0 +1,69 @@
...
+  align-items: center;
+  width: 100%;
+  padding: 64px;
+  height: 100%;
+  gap: 30px;
+}
+
+.title {
+  font-size: 80px;
+  line-height: 88px;
```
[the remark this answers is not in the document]
DiogoSoaress: You can delete the class. Why do you need the `z-index` here?

src/components/DataRoom/IndustryComparison/utils/drawDots.ts
```diff
@@ -0,0 +1,45 @@
...
+  lerpedMousePosition.x = lerp(lerpedMousePosition.x, mousePosition.x, LERP_FACTOR)
+  lerpedMousePosition.y = lerp(lerpedMousePosition.y, mousePosition.y, LERP_FACTOR)
+
+  dots.forEach((dot) => {
+    const dx = lerpedMousePosition.x - dot.x
+    const dy = lerpedMousePosition.y - dot.y
+    const distance = Math.sqrt(dx * dx + dy * dy)
+
+    // Apply scale effect
+    const scale = Math.max(1, maxScale - distance / MAX_SCALE_DISTANCE)
```
DiogoSoaress, line 37: Do you think this effect is relevant in mobile?

I even question if at all. If you are scrolling through the page, without anything to click, you don't even move the mouse 😅

src/components/DataRoom/IndustryComparison/DotGrid.tsx
```diff
@@ -0,0 +1,50 @@
...
+    if (!ctx) return
+
+    const dots = createDots(dimensions, isMobile)
+    updateCanvas(canvas, ctx, dimensions.width, dimensions.height)
+
+    let animationFrameId: number
+
+    const animate = () => {
+      if (dimensions.width > 0 && dimensions.height > 0) {
+        drawDots(ctx, dots, dimensions, mousePosition, isMobile)
```
DiogoSoaress, line 37: This function is expensive to be called non-stop. Can you only redraw the dots when necessary? (i.e, when the mouse moves or when the resolution change)

DiogoSoaress: Can you also explore a way to not have to draw the whole grid when the mouse moves but only the dots that are changing size?

src/components/DataRoom/IndustryComparison/utils/createDots.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Creates an array of dot coordinates based on given dimensions and device type.
+ * @param dimensions - The width and height of the container.
+ * @param isMobile - Boolean indicating if the device is mobile.
+ * @returns An array of objects containing x and y coordinates for each dot.
+ */
```
DiogoSoaress, line 6: The JSDOC needs to be right above the function declaration. Otherwise it is useless. You can check it by hovering any call to `createDots` in your IDE

src/components/DataRoom/IndustryComparison/styles.module.css
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  .canvasStyles {
+    margin-top: 64px;
+  }
+
+  .title {
+    font-size: 56px;
+    line-height: 60px;
+  }
```
DiogoSoaress, line 52: This is the Typography's variant `h1`. You can remove the `.title` class

src/components/DataRoom/IndustryComparison/Content.tsx
```diff
@@ -0,0 +1,38 @@
...
+  title,
+  containerRef,
+}: {
+  title: BaseBlock['title']
+  scrollYProgress: MotionValue<number>
+  containerRef: RefObject<HTMLDivElement>
+}) => {
+  const isMobile = useIsMediumScreen()
+
+  const opacityParams = [0.25, 0.35, 0.65, 0.75]
```
DiogoSoaress, line 21: Aren't these the scrollParams?

src/components/DataRoom/IndustryComparison/Content.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const opacity = useTransform(scrollYProgress, opacityParams, [0, 1, 1, 0])
```
DiogoSoaress, line 22:
```suggestion
  const opacity = useTransform(scrollYProgress, scrollParams, opacityParams)
```

src/components/DataRoom/IndustryComparison/DotGrid.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  useEffect(() => {
+    const canvas = canvasRef.current
+    if (!canvas) return
+
+    const ctx = canvas.getContext('2d')
+    if (!ctx) return
+
+    const dots = createDots(dimensions, isMobile)
+    updateCanvas(canvas, ctx, dimensions.width, dimensions.height)
```
DiogoSoaress, line 31: This is also being called unconditionally without any change in the dimensions.

src/components/DataRoom/IndustryComparison/utils/useMousePosition.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Custom hook to track mouse position or simulate it based on scroll progress.
+ * @param canvasRef - Reference to the canvas element.
+ * @param dimensions - Object containing width and height of the container.
+ * @param scrollYProgress - MotionValue for scroll progress, used on mobile devices.
+ * @returns An object with x and y coordinates representing either:
+ *          - Actual mouse position relative to the canvas (on desktop)
+ *          - Simulated position based on scroll progress (on mobile)
+ */
+
```
DiogoSoaress, line 14:
```suggestion
```

DiogoSoaress (changes requested):
Detected some causes of performance issues

src/components/DataRoom/IndustryComparison/DotGrid.tsx
DiogoSoaress: How did you address it ? `drawDots` still being called unconditionally at every possible frame and updating the canvas context for every dot

src/components/DataRoom/IndustryComparison/DotGrid.tsx
DiogoSoaress: How did you address it?

DiogoSoaress (changes requested):
There are still some performance considerations:
- The animation loop runs continuously, which can be CPU-intensive. Consider adding throttling or debouncing if the animation doesn't need to run at the highest possible frame rate.

- Canvas Updates: The canvas is updated on every frame. Ensure that `drawDots` function is optimized to minimize the amount of work done per frame.

DiogoSoaress (approved):

PR #432, opened 2024-08-02
feat(Press room): use the latest press release in the Press Room landing page

DiogoSoaress:
## What it solves
- Displays the latest press release post in the _Press room_ landing page.
- Does not display _Draft_ press releases in the page
- Fetches a `limit` of 150 blog articles on build
- Deletes the outdated `featured` field


src/components/Pressroom/PressReleases/index.tsx
```diff
@@ -23,7 +24,7 @@ const PressReleases = ({ allPosts }: { allPosts: PostEntryCollection }) => {
   const { localAllPosts } = useAllPosts(allPosts)
 
   const filteredPosts = useMemo(() => {
-    const pressPosts = localAllPosts.items.filter(isPressReleasePost)
+    const pressPosts = localAllPosts.items.filter((post) => isPressReleasePost(post) && !isDraft(post))
```
iamacook, line 27: The two instances where `isPressReleasePost` is used, now included a non-draft check. Should we this the "default", and moved into the helper (or a combined one added)?

iamacook (approved):

src/components/Pressroom/PressReleases/index.tsx
DiogoSoaress: Yes, I agree it should be combined in the helper. Good suggestion!

PR #436, opened 2024-08-07
fix(Dataroom): lazy load Framer Motion's components on client-side

### LINKED ISSUE #434: [Dataroom] fix: "Component not found" on loading the page
### Description
When refreshing or navigating to the page, sections are displaying "Component not found".

### What should be fixed
A loading state should be implemented

DiogoSoaress:
## What it solves

Resolves #434 

## How this PR fixes it
Isolate components that make use of Framer Motion in a separate component. This component is lazy loaded with `next/dynamic` and skipping SSR to avoid any layout jumps in the browser


src/components/DataRoom/AssetsSecured/SlidingContent.tsx
```diff
@@ -23,7 +23,8 @@ export default function SlidingContent({
   containerRef,
 }: Omit<BaseBlock, 'text'> & { cexes: CEX[]; containerRef: RefObject<HTMLDivElement> }) {
   const { tvlRobinhoodCEX, tvlOKX, tvlBinance, tvlSafe, lastUpdated } = useSafeDataRoomStats()
-  const { viewportWidth } = useViewportWidth()
+  const { width: viewportWidth } = useContainerSize(containerRef)
```
DiogoSoaress, line 26: `useContainerSize` hook is executing the same function using the ResizeObserver Web API

DiogoSoaress:
I identified the repetition of the Framer Motion hook to track the scroll progress but addressed the refactor in a different [PR](https://github.com/safe-global/safe-homepage/pull/443) to keep this PR scope narrow

iamacook (changes requested):
The new refs aren't being forwarded - the relative children need to also be wrapped in `forwardRef`.

DiogoSoaress:
> The new refs aren't being forwarded - the relative children need to also be wrapped in `forwardRef`.

Framer’s Motion `useScroll` is expecting to receive a `target?: RefObject$1<HTMLElement>;` instead of a `ForwardedRef<HTMLDivElement>` so I don't think that change is necessary.

iamacook (approved):

PR #438, opened 2024-08-13
feat(foundry): Hero section

DiogoSoaress:
## What it solves
Landing page Hero section

## Figma
https://www.figma.com/design/kWDCnQXTWb5xZBrNox4ny5/Safe%7BLabs%7D?node-id=1753-638&t=InkODa5bUACEYkJC-0

### DIFF
--- a/src/components/Foundry/Hero/index.tsx
+++ b/src/components/Foundry/Hero/index.tsx
@@ -12,8 +12,8 @@ const Hero = (props: BaseBlockEntry) => {
 
   const buttonsList = buttons?.filter(isEntryTypeButton) || []
 
-  const imageURL = isAsset(image) && image?.fields.file?.url ? image?.fields.file?.url : ''
-  const bgImageURL = isAsset(bgImage) && bgImage?.fields.file?.url ? bgImage?.fields.file?.url : ''
+  const imageURL = isAsset(image) && image.fields.file?.url ? image.fields.file.url : ''
+  const bgImageURL = isAsset(bgImage) && bgImage.fields.file?.url ? bgImage.fields.file.url : ''
 
   return (
     <div className={css.bg} style={{ backgroundImage: `url(${bgImageURL})` }}>
--- a/src/components/Foundry/Hero/index.tsx
+++ b/src/components/Foundry/Hero/index.tsx
@@ -1,0 +1,49 @@
+import { type BaseBlockEntry } from '@/config/types'
+import RichText from '@/components/common/RichText'
+import ButtonsWrapper from '@/components/Token/ButtonsWrapper'
+import { isAsset, isEntryTypeButton } from '@/lib/typeGuards'
+import { Container, Typography } from '@mui/material'
+import css from './styles.module.css'
+import { useIsMediumScreen } from '@/hooks/useMaxWidth'
+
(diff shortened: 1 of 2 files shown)


src/components/Foundry/Hero/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useIsMediumScreen } from '@/hooks/useMaxWidth'
+
+const Hero = (props: BaseBlockEntry) => {
+  const isMediumScreen = useIsMediumScreen()
+  const { title, text, buttons, image, bgImage } = props.fields
+
+  const buttonsList = buttons?.filter(isEntryTypeButton) || []
+
+  const imageURL = isAsset(image) && image?.fields.file?.url ? image?.fields.file?.url : ''
+  const bgImageURL = isAsset(bgImage) && bgImage?.fields.file?.url ? bgImage?.fields.file?.url : ''
```
iamacook, line 16:
```suggestion
  const imageURL = isAsset(image) && image.fields.file?.url ? image.fields.file.url : ''
  const bgImageURL = isAsset(bgImage) && bgImage.fields.file?.url ? bgImage.fields.file.url : ''
```

src/components/Foundry/Hero/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <div className={css.bg} style={{ backgroundImage: `url(${bgImageURL})` }}>
+      <div className={css.gradientHorizontal}>
+        <div className={css.spot1} />
+        <div className={css.spot2} />
+
+        {/* Networks image does not show in smaller resolutions */}
+        <div className={css.image} style={{ backgroundImage: `url(${!isMediumScreen ? imageURL : ''})` }}>
```
iamacook, line 25: You're trying to render `bgImageURL` and `imageURL` even though they can be empty strings.

src/components/Foundry/Hero/styles.module.css
```diff
@@ -0,0 +1,105 @@
+.bg {
+  background-position-x: center;
+  background-repeat: no-repeat;
+  overflow: visible;
+  background-size: cover;
+  position: relative;
+  /* margin-right: -20px; */
```
iamacook, line 7:
```suggestion
```

src/components/Foundry/Hero/index.tsx
DiogoSoaress: It is the same pattern used in https://github.com/safe-global/safe-homepage/blob/bb690f0366dbc1e397f0417c38e3a5c1ce00ad37/src/components/Token/Tokenomics/index.tsx#L39 for instance

This approach is because the types `BaseBlockEntry['bgImage']` and `BaseBlockEntry['image']` are optional even though these fields will be set in the CMS.
Because I need to check the presence of the  URL field I also need to pass a fallback.

Alternatives I see for this pattern:
- Include the fallback images in the repository code (require a deployment to change)
- create wrapper components for conditionally display the images `div`s around the children (seemed overkill knowing the images will be present)

Open to hear yet another alternative.

iamacook: Let's keep it as is then.

iamacook (approved):

PR #445, opened 2024-08-23
feat: add CSP header to all the pages

DiogoSoaress:
## What it solves
Adds an explicit Content Security Policy to the website pages

## How this PR fixes it
Sets the Content Security Policy through a `<meta>` tag

I tried to keep the allowed `src` lists as restrictive as possible without braking anything or lose of functionality


schmanu (approved):
Looks good to me.

Some rules are quite strict. If you experience that sometimes those id's or filenames change you could also allow the entire origin or the higher path as src.
e.g.
`https://cdn.contentful.com/spaces/1i5gc724wjeu/`
=> 
`https://cdn.contentful.com/spaces/`

But if the ID is always fixed, being more strict is better :)

DiogoSoaress:
> But if the ID is always fixed

Yes, in this case the Contentful space ID remains the same

PR #449, opened 2024-08-28
feat(Wallet): Vertical Slide

DiogoSoaress:
## What it solves
Implements the Vertical Slider

## Figma
https://www.figma.com/design/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?node-id=16164-3355&t=UyEbY16dBeLFg4eE-4


src/components/Wallet/VerticalSlide/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const icons = [<RecoveryIcon key="recovery" />, <ScanIcon key="scan" />, <MultipleKeysIcon key="multiple-keys" />]
+
+  const handleCardClick = (index: number) => {
+    setSelectedIndex(index)
+  }
+
+  // Change index every 5 seconds
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setSelectedIndex((prevIndex) => (prevIndex + 1) % items?.length)
```
iamacook, line 25: This will return `NaN` if `items` is undefined.

src/components/Wallet/VerticalSlide/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const VerticalSlide = ({ title, items = [] }: BaseBlock) => {
+  const [selectedIndex, setSelectedIndex] = useState(0)
+
+  const itemsImages = items.map((item) => item.image)
+  const icons = [<RecoveryIcon key="recovery" />, <ScanIcon key="scan" />, <MultipleKeysIcon key="multiple-keys" />]
+
+  const handleCardClick = (index: number) => {
+    setSelectedIndex(index)
+  }
```
iamacook, line 20: Nit: this is somewhat neglible. You can just call `setSelectedIndex` directly.

src/components/Wallet/VerticalSlide/index.tsx
DiogoSoaress: I initiate `items` to `0` in the received props

src/components/Wallet/VerticalSlide/index.tsx
DiogoSoaress: I prefer to leave the click handler

src/components/Wallet/VerticalSlide/index.tsx
iamacook: In which case, we don't need the optional chaining.

PR #476, opened 2024-10-10
feat(wallet): revamped landing page

DiogoSoaress:
## What it solves

Implements new Wallet page design

## Figma
https://www.figma.com/design/qtXXlwdslAze6yNbLLzIRa/Website-redesign-2.0?node-id=15367-15363&node-type=frame&t=horanWl7oRJbI5Vk-0

PR #487, opened 2024-11-04
chore(wallet): remove old sections

DiogoSoaress:
## What it solves
Removes the components and assets from the Wallet page that don't belong in the new design and have no reusability.


DiogoSoaress:
@iamacook Despite this clean up being part of the wallet landing page revamp, does it make sense I merge it only after merging https://github.com/safe-global/safe-homepage/pull/476 ?  The only reason I am asking this is to not bloat completely that PR with all the clean up diffs.

iamacook (approved):

PR #491, opened 2024-11-07
feat: Teaser video

DiogoSoaress:
## What it solves
Teaser landing page


src/components/Teaser/Video/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        >
+          <source src="/videos/Teaser/Teaser_no_logo.mp4" type="video/mp4" />
+          <source src="/videos/Teaser/Teaser_no_logo.webm" type="video/webm" />
+        </video>
+      </div>
+
+      {ready ? (
+        <div className={`${css.imageWrapper} ${showButton ? css.visible : ''}`}>
+          <ButtonBase target="_blank" rel="noreferrer" href={ALPHA_TELEGRAM_LINK} className={css.button}>
+            {isMobile ? <Logo className={`${css.logo} ${!showLogo ? css.hidden : css.show}`} /> : null}
```
iamacook, line 77: We can use the `mobileVideo` class here instead of `isMobile` but I don't think you can remove complete reliance on the hook so I'll leave it up to you.

iamacook (approved):

src/components/Teaser/Video/index.tsx
DiogoSoaress: This code that was using the viewport width hook was removed.

PR #496, opened 2024-11-20
feat(wallet): Dune stats

DiogoSoaress:
## What it solves
Fetches Safe numbers from Dune queries

I ended up moving the fetchers outside of the hook.


src/hooks/useSafeStats.ts
```diff
@@ -31,12 +34,25 @@ export const fetchTotalSafesDeployed = async (): Promise<number | null> => {
     .catch(() => null)
 }
 
+export const fetchMonthlyActiveUsers = async (): Promise<number | null> => {
+  return fetch(duneQueryUrlBuilder(QUERY_ID_MONTHLY_ACTIVE_USERS, DUNE_API_KEY))
+    .then((res) => res.json())
```
iamacook, line 39: The `res` may be erroneous. We should check that `res.ok` before returning the JSON.

src/hooks/useSafeStats.ts
```diff
@@ -31,12 +34,25 @@ export const fetchTotalSafesDeployed = async (): Promise<number | null> => {
...
   const formattedTotalSafesDeployed = totalSafesDeployed ? formatValue(totalSafesDeployed) : null
+  const formattedMonthlyActiveUsers = monthlyActiveUsers ? formatValue(monthlyActiveUsers) : null
 
-  return [formattedTotalTransactions, formattedTotalAssets, formattedTotalSafesDeployed]
+  return [
+    formattedTotalTransactions,
+    formattedTotalBalanceUsd,
+    formattedTotalSafesDeployed,
+    formattedMonthlyActiveUsers,
+  ]
```
iamacook, line 57: Nit: it's easy to access the wrong index. What do you think about returning an object instead?

src/pages/wallet.tsx
```diff
@@ -2,15 +2,26 @@ import type { InferGetStaticPropsType, NextPage } from 'next'
...
 
 const FAQ_CONTENT_TYPE_ID = '1jCIVFDUzFO1okK8b6TTxS'
 
 const WalletPage: NextPage<InferGetStaticPropsType<typeof getStaticProps>> = (props) => <Wallet {...props} />
 
 export async function getStaticProps() {
   const faqContent = await client.getEntry<TypeBaseBlockSkeleton>(FAQ_CONTENT_TYPE_ID)
+  const dataRoomStats = await fetchDataRoomStats()
+  const totalBalanceUsd = await fetchTotalBalanceUsd()
+  const monthlyActiveUsers = await fetchMonthlyActiveUsers()
```
iamacook, line 16: We could make these a single promise with `Promise.all`.

src/hooks/useSafeStats.ts
```diff
@@ -31,12 +34,25 @@ export const fetchTotalSafesDeployed = async (): Promise<number | null> => {
...
 }
 
+export const fetchMonthlyActiveUsers = async (): Promise<number | null> => {
+  return fetch(duneQueryUrlBuilder(QUERY_ID_MONTHLY_ACTIVE_USERS, DUNE_API_KEY))
+    .then((res) => res.json())
+    .then((data) => data.result.rows[data.result.rows.length - 1].active_users)
+    .catch(() => null)
+}
+
 export const useSafeStats = (): Array<string | null> => {
```
iamacook, line 44: Have you thought about using `swr` for this hook? We'd not need a context.

src/hooks/useSafeStats.ts
DiogoSoaress: Followed the recommendation 👍

src/hooks/useSafeStats.ts
DiogoSoaress: Added to the other fetchers as well.

PR #557, opened 2025-03-12
refactor(dataroom): add fallback values to data fetching

DiogoSoaress:
## What it solves

If the Dune query fails, the page can build anyway using fallback data as of 12/03/2025.

## How this PR fixes it

1. Includes fallback values for when `fetchDataRoomStats` returns from an error.
2. Updates component scoped fallback values.
3. Extracted the data fetcher and moved it to the `/lib` folder


src/pages/dataroom.tsx
```diff
@@ -9,49 +30,58 @@ const DataRoomPage: NextPage<InferGetStaticPropsType<typeof getStaticProps>> = (
...
+        tvlOKX: Number(cex_tvl_okx),
+        tvlRobinhoodCEX: Number(cex_tvl_robinhood),
+        tvlSafe: Number(safe_tvl),
         tvlUniswap: Number(defi_tvl_uniswap),
-        lastUpdated: Number(last_updated),
-        annualSwapFees: Number(annual_swap_fees),
-        annualStakeFees: Number(annual_stake_fees),
-        annualisedOutgoingTVP: annualised_outgoing_tvp,
+        tvpToGDPPercentage: tvp_perc_world_gdp,
+        usdcPercentageStored: total_usdc_share,
```
DiogoSoaress, line 84: Took the chance to alphabetically order the properties

src/components/DataRoom/IndustryComparison/index.tsx
```diff
@@ -5,7 +5,7 @@ import css from './styles.module.css'
 
 const Content = dynamic(() => import('./Content'))
 
-export const LAST_UPDATED_FALLBACK = 1722946836.34
+export const LAST_UPDATED_FALLBACK = 1740577705.913 // 26.02.2025
```
iamacook, line 8: Nit: as we added `_` separators elsewhere, let's do so here as well.

src/pages/dataroom.tsx
```diff
@@ -9,49 +30,58 @@ const DataRoomPage: NextPage<InferGetStaticPropsType<typeof getStaticProps>> = (
...
+    tvp_perc_world_gdp,
+    // usdc_in_safes,
   } = dataRoomStats
 
   return {
     props: {
       safeDataRoomStats: {
-        tvpToGDPPercentage: tvp_perc_world_gdp,
-        usdcPercentageStored: total_usdc_share,
-        // This is a temp workaround. The received value should be a number.
```
iamacook, line 39: Is this no longer the case?

src/components/DataRoom/IndustryComparison/index.tsx
DiogoSoaress: It is a timestamp. No need there I'd say

src/pages/dataroom.tsx
DiogoSoaress: (sorry) I ordered the props alphabetically. These two were moved to the end.

DiogoSoaress: I decided it is not relevant anymore. It is fine that the received values from the Dune query are either strings or numbers.

iamacook (approved):

